PR #5, opened 2026-07-11
Stage 3 — MPI outer-k decomposition for run_eph_over_k_and_kq (FMP)

jaemolihm:
## Stage 3 of the GPU-port merge plan (FMP-specific)

Cherry-picks the two genuinely FMP-unique commits from \`gpu-fmp-speedup\` onto the post-Stage-2b \`main\`, after subtracting the Stage-2b duplicates.

| commit | change |
|---|---|
| \`72fb148\` (was \`f1e42ed\`) | feat: MPI outer-k decomposition for \`run_eph_over_k_and_kq\` (multi-CPU/GPU) |
| \`3304d2d\` (was \`0024017\`) | fix: \`kpoints_grid_range\` uses BZ-fraction weight (correct under MPI k-split) |

### What it does
- \`mpi_comm_k\` now splits the **outer** k-points across ranks; each rank computes its k-slice's coupling (rank-local \`calc.g2[:, i_local, :]\`). k+q / q-grid / phonons stay full per rank. CPU and GPU inner loops unchanged.
- Guards \`mpi_comm_k\` to \`el_kq_from_unfolding = false\` (unfolding path is not split-aware); \`mpi_comm_q\` still errors (not implemented).
- Fixes latent bugs in the symmetry+MPI(+GPU) path (never exercised before, no driver passed \`mpi_comm_k\`): \`filter_kpoints(::MPI.Comm)\` missing \`use_gpu\`; \`shift\` \`Vector\` default spuriously tripping the symmetry guard; \`kpoints_grid(::MPI.Comm; symmetry)\` scattering a \`GridKpoints\`.
- Weight fix: \`kpoints_grid_range\` weights each point by the BZ fraction \`1/prod(ngrid)\` (not \`1/length(rng)\`), so a per-rank sub-range's Σ w_k is correct. Serial/full-grid result is identical.

### Reconciliation vs the plan
Both commits cherry-picked **clean** onto \`main\` (no conflicts). Verified the merged \`main..HEAD\` diff is content-identical to the original \`f1e42ed~1..0024017\` range (only blob-hash header lines differ, from unrelated Stage-2b edits elsewhere in the same files).

### Serial path unchanged
Default \`mpi_comm_k = nothing\`: the new guards are no-ops and \`kpoints_grid_range\`'s full-range weight is unchanged, so the standard (serial, single-GPU) test suite is the no-regression guardrail.

Draft until the pre-commit gate (critical review + GPU tests + benchmark→PERF_LOG) passes.

### DIFF
--- a/src/calculator/run_eph_over_k_and_kq.jl
+++ b/src/calculator/run_eph_over_k_and_kq.jl
@@ -481,14 +481,18 @@ end
 #  batched over k-batches and q-batches with device staging — differs from the per-(k,q) CPU loop,
 #  not because it holds any device-specific code.
 #
-#  Minimal first step — the rest is asserted off (not implemented on the GPU path):
-#    * no polar / long-range terms (so ϵs = 1, no dipole correction)
-#    * full bands, no energy window  ->  nband == nw at every k / k+q (uniform batch shapes)
-#    * commensurate k / k+q grids (precompute_ph) so phonon states are precomputed
-#    * no covariant derivative of g, no screening, no el_kq_from_unfolding,
-#      and energy_conservation = (:None, 0.0)
-#  (IBZ outer-k symmetry and outer-k MPI decomposition ARE supported — both are handled upstream
-#   in the shared `_setup`, so this loop stays symmetry- and MPI-agnostic.)
+#  Supported (all handled upstream in the shared `_setup`, so this loop itself is agnostic to them):
+#    * energy windows (window_k / window_kq): the k side carries only its in-window bands, the
+#      window is applied in the calculator scatter
+#    * IBZ outer-k symmetry
+#    * outer-k MPI decomposition (mpi_comm_k)
+#  Not supported — asserted off on the GPU path (see the scope-assert block below):
+#    * polar / long-range terms
+#    * incommensurate k / k+q grids: requires precompute_ph (phonon states precomputed)
+#    * covariant derivative of g
+#    * screening
+#    * el_kq_from_unfolding (directly-computed k+q only)
+#    * energy_conservation other than (:None, 0.0)
 #
 #  Buffer reuse: device buffers are allocated once before the k loop and reused for every
 #  (k, q), so the loop itself allocates almost nothing.
--- a/src/calculator/run_eph_over_k_and_kq.jl
+++ b/src/calculator/run_eph_over_k_and_kq.jl
@@ -32,13 +32,11 @@ function run_eph_over_k_and_kq(
     end
 
     # Outer-k MPI decomposition (multi-CPU/GPU): `mpi_comm_k` splits the OUTER k-points across ranks
-    # — each rank computes the e-ph coupling only for its k-slice, yielding a rank-local
-    # `calc.g2[:, i_local, :]` / `el_i`. k+q, the q-grid, and the phonon states stay FULL on every
-    # rank (so any k can scatter to any k+q). `filter_kpoints` does the split + load-balance in the
-    # shared `_setup`; the EliashbergCalculator is already rank-local; and the CPU and GPU inner
-    # loops are UNCHANGED (they iterate whatever `kpts` they are handed). The solver then decomposes
-    # over the matching outer states i (see MigdalEliashberg). q-decomposition (mpi_comm_q) is a
-    # separate, not-yet-implemented scheme. See GPU_PROGRESS.md "Multi-GPU / MPI".
+    # — each rank computes the e-ph coupling for its k-slice only (rank-local `calc.g2` / `el_i`),
+    # while k+q, the q-grid, and the phonon states stay FULL per rank (so any k can scatter to any
+    # k+q). `filter_kpoints` does the split + load-balance in `_setup`; the CPU and GPU inner loops
+    # are unchanged (they iterate whatever `kpts` they receive). `mpi_comm_q` is a separate scheme,
+    # not yet implemented.
     mpi_comm_q === nothing || error("mpi_comm_q not implemented (use mpi_comm_k for outer-k decomposition)")
     (mpi_comm_k === nothing || !el_kq_from_unfolding) || throw(ArgumentError(
         "mpi_comm_k requires el_kq_from_unfolding = false (k+q stays full per rank; the unfolding " *
@@ -487,8 +485,10 @@ end
 #    * no polar / long-range terms (so ϵs = 1, no dipole correction)
 #    * full bands, no energy window  ->  nband == nw at every k / k+q (uniform batch shapes)
 #    * commensurate k / k+q grids (precompute_ph) so phonon states are precomputed
-#    * no covariant derivative of g, no screening, no symmetry/unfolding, single node (no MPI),
+#    * no covariant derivative of g, no screening, no el_kq_from_unfolding,
 #      and energy_conservation = (:None, 0.0)
+#  (IBZ outer-k symmetry and outer-k MPI decomposition ARE supported — both are handled upstream
+#   in the shared `_setup`, so this loop stays symmetry- and MPI-agnostic.)
 #
 #  Buffer reuse: device buffers are allocated once before the k loop and reused for every
 #  (k, q), so the loop itself allocates almost nothing.
--- a/src/calculator/run_eph_over_k_and_kq.jl
+++ b/src/calculator/run_eph_over_k_and_kq.jl
@@ -31,8 +31,18 @@ function run_eph_over_k_and_kq(
         end
     end
 
-    mpi_comm_k === nothing || error("mpi_comm_k not implemented")
-    mpi_comm_q === nothing || error("mpi_comm_q not implemented")
+    # Outer-k MPI decomposition (multi-CPU/GPU): `mpi_comm_k` splits the OUTER k-points across ranks
+    # — each rank computes the e-ph coupling only for its k-slice, yielding a rank-local
+    # `calc.g2[:, i_local, :]` / `el_i`. k+q, the q-grid, and the phonon states stay FULL on every
+    # rank (so any k can scatter to any k+q). `filter_kpoints` does the split + load-balance in the
+    # shared `_setup`; the EliashbergCalculator is already rank-local; and the CPU and GPU inner
+    # loops are UNCHANGED (they iterate whatever `kpts` they are handed). The solver then decomposes
+    # over the matching outer states i (see MigdalEliashberg). q-decomposition (mpi_comm_q) is a
+    # separate, not-yet-implemented scheme. See GPU_PROGRESS.md "Multi-GPU / MPI".
+    mpi_comm_q === nothing || error("mpi_comm_q not implemented (use mpi_comm_k for outer-k decomposition)")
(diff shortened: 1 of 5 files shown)


jaemolihm:
## Pre-commit gate: PASSED ✅

**1. Critical review** (`critical-review` subagent, reviewer persona over the full diff) — verdict **"ship it"**. No critical/correctness issues. The MPI split, the three latent-bug fixes, and the BZ-fraction weight change are all correct; the serial (`mpi_comm_k = nothing`) path is provably untouched. Findings were comment/convention cleanups, applied in `8597488`:
- dropped external-calculator names (`MigdalEliashberg`/`EliashbergCalculator`) from core-library comments (library is calculator-agnostic);
- dropped the untracked-`GPU_PROGRESS.md` reference; shortened the outer-k MPI block;
- fixed the now-stale GPU-loop scope comment (`"no symmetry/unfolding, single node (no MPI)"` → IBZ outer-k symmetry **and** outer-k MPI are both supported, handled upstream in `_setup`).

**Open question (not a diff defect, deliberately not changed):** the CPU path does not guard `mpi_comm_k` + `screening_params` together (the GPU path already asserts `screening_params === nothing`). If screening sums over the outer-k BZ, it would need an MPI reduction under an outer-k split. No current consumer combines the two; left as-is per "ask before touching physics-adjacent code."

**2. GPU tests** — **81/81 green** on RTX A6000 / `ccqlin059`. Only comments changed after this run, so the executable tree is identical to the tested one. (Suite has no `mpi_comm_k` coverage; the np=2-vs-serial MPI validation is in the `f1e42ed` commit message.)

**3. Benchmark → PERF_LOG** (`25ee979`) — single-GPU no-regression check (`bench_eliashberg_loop_gpu.jl`, window ±0.3 eV):

| grid | n_i | CPU | GPU | vs Stage 2b GPU |
|-|-|-|-|-|
| 16³ | 432  | 0.59 s | 0.54 s | 0.55 s |
| 24³ | 1716 | 6.85 s | 3.00 s | 2.94 s |
| 32³ | 4052 | 38.7 s | 8.96 s | 8.83 s |

32³ Δ≈1.5% (run-to-run noise) — **no regression**, as expected since Stage 3 leaves the serial inner loop unchanged.

Ready for the human review round. Still draft until sign-off.

src/common/kpoints.jl
```diff
@@ -106,9 +106,12 @@ function kpoints_grid(ngrid, mpi_comm::Union{MPI.Comm, Nothing}=nothing; shift=(
...
         end
         if mpi_comm isa MPI.Comm
-            # Create the irreducible k points in the root. Then redistribute.
+            # Create the irreducible k points in the root. Then redistribute. Reduce the root's
+            # `GridKpoints` to a plain `Kpoints` first so every rank holds the same type for the
+            # collective `mpi_scatter` (which has a `Kpoints` method) and returns `Kpoints` — the
+            # same type the non-symmetry MPI branch (`kpoints_grid_range`) returns.
             kpoints = if mpi_isroot(mpi_comm)
-                kpoints_grid_symmetry(ngrid, symmetry; ignore_time_reversal)
+                Kpoints(kpoints_grid_symmetry(ngrid, symmetry; ignore_time_reversal))
```
jaemolihm: Does this mean if using MPI the return of this function is Kpoints not GridKpoints ?

Should we define bcast for GridKpoints ?

src/common/kpoints.jl
```diff
@@ -146,8 +149,12 @@ function kpoints_grid_range(ngrid::NTuple{3, Int}, rng::UnitRange{Int}, ::Type{F
         i = mod(div(ik-1 - k - j*nk3, nk2*nk3), nk1)
         push!(kvecs, Vec3{FT}(i/nk1, j/nk2, k/nk3) .+ shift)
     end
-    nk = length(kvecs)
-    Kpoints(nk, kvecs, fill(1/nk, (nk,)), (nk1, nk2, nk3))
+    # BZ weight is the grid fraction 1/prod(ngrid) per point — NOT 1/length(rng). They coincide for
```
jaemolihm: nrange -> nk, this is simple bugfix, no comment needed, but make sure this is tested

src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -477,8 +485,10 @@ end
 #    * no polar / long-range terms (so ϵs = 1, no dipole correction)
 #    * full bands, no energy window  ->  nband == nw at every k / k+q (uniform batch shapes)
 #    * commensurate k / k+q grids (precompute_ph) so phonon states are precomputed
-#    * no covariant derivative of g, no screening, no symmetry/unfolding, single node (no MPI),
+#    * no covariant derivative of g, no screening, no el_kq_from_unfolding,
 #      and energy_conservation = (:None, 0.0)
+#  (IBZ outer-k symmetry and outer-k MPI decomposition ARE supported — both are handled upstream
```
jaemolihm: Do a list of supported options

src/common/kpoints.jl
jaemolihm: Done in `9f292e6`. Yes — before this, the MPI symmetry branch reduced to `Kpoints` before scattering, so it returned `Kpoints` while the serial branch returned `GridKpoints`. Per your follow-up, I now keep `GridKpoints`: added `mpi_scatter`/`mpi_gather`/`mpi_gather_and_scatter` methods for `GridKpoints` (each rank's scattered subset is grid-aligned and rebuilds its own xk→ik hash) plus an empty `GridKpoints{T}()` constructor for the non-root receive placeholder. So both the serial and MPI symmetry branches now return `GridKpoints`, and `filter_kpoints(::MPI.Comm)`'s load-balancing gather+scatter keeps it too. Added a return-type testset (COMM_SELF) covering `GridKpoints`→`GridKpoints`. Left a TODO on the non-symmetry branch (`kpoints_grid_range`) to also return `GridKpoints`, per your note.

src/common/kpoints.jl
jaemolihm: Done in `9f292e6`. Reverted `nrange`→`nk` and dropped the comment (kept the `1/prod(ngrid)` weight). Added tests in `test/test_kpoints.jl` ("kpoints: grid"): a full range weights each point `1/prod(ngrid)`; a strict sub-range (`3:9`) carries the same global BZ fraction (not `1/7`); and a partition of `1:nk` into sub-ranges has weights summing to 1.

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: Done in `9f292e6`. Rewrote the GPU-loop scope comment as explicit **Supported** / **Not supported** lists. Supported (all handled upstream in `_setup`): energy windows (`window_k`/`window_kq`), IBZ outer-k symmetry, outer-k MPI (`mpi_comm_k`). Not supported / asserted off: polar/long-range terms, incommensurate grids (needs `precompute_ph`), covariant derivative of g, screening, `el_kq_from_unfolding`, and `energy_conservation != (:None, 0.0)`. (This also corrected the stale "full bands, no energy window" claim — windowing has been supported since the Stage-1/2b k-side projection.)

src/common/kpoints.jl
```diff
@@ -376,10 +376,41 @@ end
...
+the result stays a `GridKpoints` — matching the serial symmetry return type.
+"""
+function mpi_scatter(k::GridKpoints{FT}, comm::MPI.Comm) where {FT}
+    ngrid = mpi_bcast(k.ngrid, comm)
+    vectors = mpi_scatter(k.vectors, comm)
+    weights = mpi_scatter(k.weights, comm)
+    GridKpoints(Kpoints{FT}(length(vectors), vectors, weights, ngrid), ngrid)
+end
+
+function mpi_gather(k::GridKpoints{FT}, comm::MPI.Comm) where {FT}
```
jaemolihm, line 399: Should we have a equality check for all ngrid ?

src/common/kpoints.jl
```diff
@@ -376,10 +376,41 @@ end
...
+    weights = mpi_gather(k.weights, comm)
+    new_ngrid = _gather_ngrid(k.ngrid, comm)
+    if mpi_isroot(comm)
+        GridKpoints(Kpoints{FT}(length(kvectors), kvectors, weights, new_ngrid), new_ngrid)
+    else
+        GridKpoints{FT}()
+    end
+end
+
+"Collect and uniformly redistribute GridKpoints among processers (each even chunk stays grid-aligned)."
```
jaemolihm: What does :(each even chunk stays grid-aligned): mean 

And why is mpi_gather_and_scatter needed? (Add 1-2 sentence)

test/test_kpoints.jl
```diff
@@ -174,3 +186,34 @@ end
     single = GridKpoints(kpoints_grid((1, 1, 1)))
     @test_throws ArgumentError combine_kpoint_grids(single, single, +, (10^7, 10^7, 10^7))
 end
+
+@testset "kpoints: mpi_scatter return type" begin
```
jaemolihm, line 190: add comment TODO: actualy test with mpi

src/common/kpoints.jl
jaemolihm: It's already handled by `_gather_ngrid` (shared with the `Kpoints` gather): it broadcasts the root's ngrid and does `mpi_reduce(ngrid == ngrid_root, &, comm)`, returning `(0,0,0)` if they differ. For `GridKpoints` that `(0,0,0)` then fails loud in the constructor (`all(ngrid .> 0)` throws) — so an inconsistent grid can't silently produce a wrong result. In the actual usage (outer-k MPI split) every rank shares the same k-grid, so this never triggers. The only downside is that the fail-loud message would be the generic `ngrid must be set or provided` rather than an explicit "ngrids differ across ranks". I can add an explicit equality assertion for a clearer error if you'd prefer — leaving this open for you to decide.

src/common/kpoints.jl
jaemolihm: Clarified in `cc61cfa`. "Grid-aligned" = every k-point is a node of the original regular grid, so any redistribution of them is still made of grid nodes and each rank's chunk is again a valid `GridKpoints`. Why it's needed: `filter_kpoints` drops out-of-window points, which leaves an uneven count per rank; `mpi_gather_and_scatter` gathers all survivors to the root and re-scatters them evenly to restore load balance. Expanded the comment to say both.

test/test_kpoints.jl
jaemolihm: Added in `cc61cfa`: a comment noting COMM_SELF exercises only the single-rank path, plus `TODO: actually test with a real multi-rank MPI communicator (np > 1)`.

src/common/kpoints.jl
jaemolihm: Resolving as discussed — keeping the existing `_gather_ngrid` handling (mismatched grids fail loud in the GridKpoints constructor); no explicit assertion added. Can revisit if a clearer error message is wanted later.

jaemolihm:
Merged to `main` as `439c82d` (squash, local git merge — consistent with Stages 1/2a/2b). All review threads resolved; gate passed (critical review + 81/81 GPU tests on RTX A6000 + benchmark in PERF_LOG, no regression). Closing.

PR #10, opened 2026-07-23
Multigrid (non-uniform) BTE k-sampling via StateSelection; unify filtering into filter_electron_states

jaemolihm:
## Motivation

The linearized BTE stores a dense scattering matrix `Sᵢ` (`n_i × n_f` per T) that grows ∝ nk⁶ — uniform Cu nk=200/±0.8 needs ~733 GB and OOMs. But near-μ quantities (especially the **Seebeck coefficient**, a Mott derivative) need dense k-sampling *right at μ*, where a uniform fine grid is infeasible. **Multigrid sampling** — a fine grid in a narrow window near μ merged with a coarse grid over a wide window — shrinks both `n_i` and `n_f`, recovering near-μ accuracy at a fraction of the memory.

## Change

The multigrid is represented as a **state selection** (per-`(k, band)` states), not a merged k-point set, and k/band filtering is unified into one primitive.

- **`AbstractBandStates`** (`src/band_states.jl`): new abstract supertype. `BandStates` (rich, +`es`/`vs`) and a new lean **`StateSelection`** — the selected `(k, band)` pairs + per-state weights + `nstates_base` + per-k band extent — both subtype it; the shared `(k,band)`-selection machinery (`state_index`, `indmap`, `state_weights`, unfolding, device map) dispatches on the abstract type. Adds `band_range` and `unfold_band_states`.
- **`filter_electron_states`** (`src/filter.jl`): the single filtering primitive, returns a `StateSelection` (single-pass MPI-aware construction). `filter_kpoints` is kept as a deprecated tuple-returning alias. **`filter_electron_states_multigrid`** (renamed from `filter_kpoints_multigrid`) emits per-`(k,band)` **clean-partition** double-grid weights — fine-node narrow-window bands at the fine weight, wide-only bands at the coarse weight. This fixes the fine-node wide-tail **band over-selection** (a correctness bug) and is provably equal to the corrected double-grid estimator.
- **`run_eph_over_k_and_kq`**: consumes a prebuilt `StateSelection` **as-is** (no internal filter/unfold on that path); a grid argument runs the primitive as sugar. The `nelec_below_window_k/kq` override kwargs are **removed** — `nstates_base` rides on the selection and the auto-μ solve reads it. The caller builds the full-BZ k+q selection explicitly via `unfold_band_states`.
- **Per-final-state scattering weight** (CPU + GPU) replaces the per-k+q-point weight; it reduces to the old weight when per-state weights are empty, so uniform-grid results are bit-identical.
- **`compute_electron_states`** consumes the selection (per-k band extent, via `set_window!`). The sibling drivers (`run_eph_over_k_and_q`, `run_eph_over_q_and_k`) forward `sel_k` so their calculators receive the below-window carrier count.
- Worked example: `examples/bte_multigrid.jl`.

## Test

- `test/test_filter_electron_states.jl`: serial == MPI state set + `nstates_base` + `band_range` (real 2-rank `mpiexec` check for the band-range redistribution); `shift` round-trip.
- `test/boltzmann/test_multigrid_weights.jl`: e-ph-independent per-**band** BZ quadrature (`Σ` per-state weight over each band ≈ 1).
- `test/boltzmann/test_multigrid_transport.jl`: multigrid σ vs uniform reference + CPU-vs-GPU (~1e-13) + auto-μ success + a wide-tail-band over-selection regression guard.
- Validation (Pb, GPU): multigrid 12³/±0.1 + 6³/±0.4 vs uniform 12³/±0.4 → σ reldiff ~1.9%, vs ~140% for coarse-alone.

## Notes

- **Uniform-grid results are bit-identical** to before (the per-final-state weight reduces to the per-k+q-point weight when per-state weights are empty).
- `band_range` returns the true lowest in-window band; the old `filter_kpoints` `band_min` was `0` whenever any k was filtered (a zero-init artifact). Computed conductivity is unchanged; only `iband_min`/`nband` output metadata moves.
- **Draft — for review.** Design + resolved decisions live in `plans/multigrid_bandstates.md` and `plans/filter_electron_states_unification.md` (intentionally not committed).


examples/bte_multigrid.jl
```diff
@@ -0,0 +1,96 @@
...
+
+"""
+    run_bte_multigrid(model, nk_fine, nk_coarse; η, window_fine, window_wide, occ,
+                      method, use_gpu, symmetry)
+
+Run one multigrid BTE transport calculation and return the SERTA and full-BTE conductivity
+tensors (SI, `(Ω·cm)⁻¹`, shape `(3,3,nT)`) plus the calculator. `nk_fine` must be a multiple of
+`nk_coarse`. Builds a single multigrid spec shared by the k and k+q arguments.
+"""
+function run_bte_multigrid(model, nk_fine, nk_coarse; η, window_fine, window_wide, occ,
```
jaemolihm: window_fine -> window_narrow (because window_wide)
Also nk_fine, nk_coarse -> nk_narrow, nk_wide

examples/bte_multigrid.jl
```diff
@@ -0,0 +1,96 @@
...
+        method = 5, use_gpu = true, symmetry = model.symmetry)
+    kmg, nelec_below_window = EP.filter_kpoints_multigrid(
+        (nk_fine, nk_fine, nk_fine), (nk_coarse, nk_coarse, nk_coarse),
+        window_fine, window_wide, model.nw, model.el_ham; symmetry, use_gpu)
+
+    calc = BoltzmannCalculator{Float64}(; occ,
+        smearing_list = [SmearingType(:Lorentzian, η) for _ in 1:length(occ)],
+        occupation_method = method)
+    EP.run_eph_over_k_and_kq(model, kmg, kmg;
```
jaemolihm: 1. Isn't passing kmg and window_wide problematic?

Suppose there is a nk_fine mesh k point where band 1 is in window_fine and 2 is in window_wide. This k point is selected and inside kmg. But, for the electron state, it is selected by window_k = window_wide, so both band 1 and 2 are selected. We only want band 1 to be selected,

2. Is it okay to pass irreducible k+q points as kqpts ? Is it unfolded inside?

src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -23,6 +23,12 @@ function run_eph_over_k_and_kq(
         # see outer k SERIALLY (one `OuterIteration` bracket + payload per k); this only sets the
         # interpolation-staging width. No deprecated alias for the former name `nk_batch_max`.
         nk_outer_batch_max = 256,
+        # Override the below-window carrier counts (from `filter_kpoints_multigrid`) instead of
+        # recomputing them by filtering the input grid. Required for a pre-built WINDOWED grid (a
+        # multigrid), whose below-window k-points are absent so the recompute under-counts and the
+        # chemical-potential solve fails to bracket. `nothing` = recompute as usual.
+        nelec_below_window_k = nothing,
```
jaemolihm: This, plus the band issue I mentioned before, indicates that passing kpts is not a good abstraction. We need an information on "All states (band, kpoint) selected plus the nelec below window". Maybe we pass BandStates ...?

examples/bte_multigrid.jl
jaemolihm: Done in 0fc38f4: renamed `window_fine → window_narrow`, `nk_fine → nk_narrow`, `nk_coarse → nk_wide` in the example (windows named by width; each mesh-size var named for the window it fills).

examples/bte_multigrid.jl
jaemolihm: On (2), yes: with symmetry set, `_setup_eph_over_k_and_kq` treats the passed kqpts as IBZ points, window-filters them, then `unfold_kpoints` expands to the full BZ. You noted you'd rather a passed kqpts be used as-is — agreed; folding that into the redesign below (caller builds the full selection; no implicit internal re-filter/unfold).

On (1) — the band over-selection — you confirmed it's a real correctness bug. Both (1) and thread [run_eph_over_k_and_kq.jl:30] are being addressed together via the BandStates redesign; EPjl-planner is drafting `plans/multigrid_bandstates.md` now. Leaving this thread open pending that plan.

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: Agreed — this and the band-selection issue both stem from collapsing the multigrid into a plain k-set and re-filtering by a single window. `BandStates` (src/band_states.jl) already carries per-(k,band) states plus `nstates_base` (below-window carrier count), so it's the natural replacement and removes the `nelec_below_window_k/kq` override. Being planned in `plans/multigrid_bandstates.md` (EPjl-planner). Leaving open pending the plan.

examples/bte_multigrid.jl
jaemolihm: Both addressed in c43d2ea.

**Band over-selection (1):** the multigrid now emits a `StateSelection` with per-(k,band) window tagging — at a fine-only node only the narrow-window bands (fine weight), at a coincident fine∩coarse node the narrow bands at fine weight AND the wide-only bands at coarse weight. So a fine node no longer carries its wide-tail bands at the fine weight. (The clean-partition rule was verified equal to the corrected double-grid counter-term.)

**k+q as-is (2):** the k+q side is now a `StateSelection` built explicitly by the caller via `unfold_band_states(sel_k, symmetry)` and consumed as-is — no hidden internal `filter_kpoints` + `unfold_kpoints` on the selection path.

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: Done in c43d2ea. The k-set abstraction is replaced: a new `abstract type AbstractBandStates` with a lean `StateSelection` carrying the selected (k, band) pairs + per-state weights + `nstates_base` (the below-window count). `run_eph_over_k_and_kq` consumes a prebuilt selection directly, so the `nelec_below_window_k/kq` override kwargs are gone (the count rides on the selection). Filtering is unified into a single `filter_electron_states` primitive that returns a `StateSelection` (`filter_kpoints` kept as a deprecated tuple-returning alias).

ext/ElectronPhononCUDAExt.jl
```diff
@@ -319,7 +319,7 @@ function _bte_window_accumulate_kernel!(Sₒ_out, Sᵢ_out, g2vals, ωqmat, imap
         ikq = ikqs[j]              # k+q point of this q within the batch
         f = imap_f[m, ikq]         # inner (k+q) state index; 0 = out of window → skip
         f > 0 || return
-        ek = e_i[i]; ekq = e_f[f]; wtq = wq[ikq]
+        ek = e_i[i]; ekq = e_f[f]; wtq = wf[f]   # per-final-state weight ([DECISION 3])
```
jaemolihm: Don't have comment like `DECISION 3` in the code. This is not useful for future readers who don't have context of the plan. Do a full audit of current PR and remove such comments.

Update the EPjl reviewer agent to flag such comments.

src/boltzmann/boltzmann_calculator.jl
```diff
@@ -128,22 +136,27 @@ function setup_calculator!(calc::BoltzmannCalculator{FT}, kpts, qpts, el_states;
...
+    # selection (the below-window count rides on the selection) and brackets normally. `bte_compute_μ!`
+    # reads BTStates fields directly; the μ formula is unchanged.
     if !chemical_potential_is_computed(calc.occ)
-        el, _ = electron_states_to_BTStates(el_states, kpts, nelec_below_window_k)
-        bte_compute_μ!(calc.occ, el; do_print=true)
+        el_bt = BTStates{FT}(; n = calc.el_i.n, nk = calc.el_i.kpts.n, nband = calc.el_i.nband,
+            e = calc.el_i.es, vdiag = calc.el_i.vs, k_weight = state_weights(calc.el_i),
+            xks = state_xks(calc.el_i), iband = calc.el_i.ibands, ngrid = calc.el_i.kpts.ngrid,
+            nstates_base = calc.el_i.nstates_base)
+        bte_compute_μ!(calc.occ, el_bt; do_print=true)
```
jaemolihm: Why not pass calc.el_i ?

src/boltzmann/boltzmann_calculator.jl
```diff
@@ -207,14 +220,17 @@ function run_calculator!(calc::BoltzmannCalculator{FT}, p::EPData, ctx::LoopCont
...
-    (; el_k, el_kq, ph, wtq) = epdata
+    (; el_k, el_kq, ph) = epdata
     method = calc.occupation_method
     @inbounds for n in el_k.rng
         ek = el_k.e[n]
         for m in el_kq.rng
             ind_el_f = state_index(calc.el_f, ikq, m)
             ind_el_f == 0 && continue
             ekq = el_kq.e[m]
+            # Per-final-state weight ([DECISION 3]): the inner state f = ind_el_f carries its own
```
jaemolihm: Remove this comment

src/boltzmann/boltzmann_calculator.jl
```diff
@@ -207,14 +220,17 @@ function run_calculator!(calc::BoltzmannCalculator{FT}, p::EPData, ctx::LoopCont
...
+            # (fine or coarse) BZ weight, replacing the per-k+q-point wtq.
+            wtq = calc.w_f[ind_el_f]
```
jaemolihm: calc.el_f[ind_el_f].weight or something like this? Then calc don't need explicit w_t (it is avaialble via el_f anyway)

src/calculator/eph_setup_common.jl
```diff
@@ -15,13 +15,19 @@ function _setup_electron_k(
...
     (; nw) = model
-    kpts, iband_min, iband_max, nelec_below_window_k = maybe_time(verbosity) do
-        filter_kpoints(kpts_input, nw, model.el_ham, window_k, mpi_comm_k; symmetry, fourier_mode, use_gpu)
+    sel = maybe_time(verbosity) do
+        filter_electron_states(kpts_input, nw, model.el_ham, window_k; symmetry, fourier_mode, use_gpu, mpi_comm=mpi_comm_k)
     end
+    kpts = sel.kpts
+    br = band_range(sel)
+    iband_min, iband_max = first(br), last(br)
+    # State computation stays window-based (per [DECISION 2]): no per-k band extent from the selection.
```
jaemolihm: Why? Why not pass sel ?

src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -1,7 +1,7 @@
 function run_eph_over_k_and_kq(
         model       :: Model{FT},
-        kpts_input  :: Union{NTuple{3,Int}, Kpoints, GridKpoints},
-        kqpts_input :: Union{NTuple{3,Int}, Kpoints, GridKpoints},
+        kpts_input  :: Union{NTuple{3,Int}, Kpoints, GridKpoints, StateSelection},
```
jaemolihm: Maybe call StateSelection -> FilteredStates ? Then it aligns nicely with filter_* function and AbstractBandStates name.

src/calculator/run_eph_over_q_and_k.jl
```diff
@@ -151,7 +151,7 @@ function _setup_eph_over_q_and_k(
     # setup cost after the eigensolve). Default is the conservative full list.
     el_k_quantities = isempty(calculators) ? ["eigenvalue", "eigenvector", "velocity", "position"] :
         unique(reduce(vcat, required_el_k_quantities(c) for c in calculators))
-    (; kpts, iband_min, iband_max, nelec_below_window_k, el_k_save) = _setup_electron_k(
+    (; kpts, iband_min, iband_max, el_k_save, sel) = _setup_electron_k(
```
jaemolihm: Make _setup_electron_k return sel_k not sel

src/compute_states.jl
```diff
@@ -30,6 +30,36 @@ function compute_electron_states(model::Model{FT}, kpts, quantities, window=(-In
...
+        _compute_electron_states_gpu!(states, model, kpts, quantities, sel.band_extent)
+    else
+        _compute_electron_states_cpu!(states, model, kpts, quantities, sel.band_extent; fourier_mode)
+    end
+    states
+end
+
+# The per-k band window for `set_window!`: an energy-window tuple applies to every k; a per-k vector
+# of band ranges (from a `StateSelection`) is indexed by k.
+_window_for(window::Tuple, ik) = window
```
jaemolihm, line 62: Why is this needed ?

src/calculator/eph_setup_common.jl
jaemolihm: Done in 711b5fd. _setup_electron_k now computes states via compute_electron_states(model, sel_k, ...) using the selection per-k band extent. Behavior-identical to the window form in the normal case, and it makes the computed states exactly consistent with the selection (removes a potential filter-vs-compute band-range mismatch at window edges).

ext/ElectronPhononCUDAExt.jl
jaemolihm: Done in 711b5fd. Swept all [DECISION N] and plans/*.md references out of the comments this PR introduced (src + tests), keeping only self-contained explanations; pre-existing DECISION refs from earlier merged work were left untouched. Also added a plan-reference-comment check to the EPjl-reviewer agent so this gets flagged going forward.

src/boltzmann/boltzmann_calculator.jl
jaemolihm: Done in 711b5fd. Dropped the calc.w_f field; the per-final-state weight now rides on el_f.weights (materialized), read as an O(1) field reference in the CPU loop and for the GPU dev.wf buffer. Scattering stays bit-identical (CPU == GPU, uniform and multigrid).

src/compute_states.jl
jaemolihm: Kept, with a self-contained comment added in 711b5fd. _window_for lets compute_electron_states accept either an energy-window Tuple (uniform; still used by the k+q / window callers) or a per-k Vector of band ranges (from a FilteredStates selection), dispatching between the two.

src/boltzmann/boltzmann_calculator.jl
jaemolihm: Done in 711b5fd. bte_compute_mu! now accepts the BandStates directly (via the AbstractBandStates accessors es / bt_weights / ibands / nstates_base); the throwaway temp BTStates is gone. mu is bit-identical (delta = 0.0) and the legacy BTStates callers are unchanged.

src/boltzmann/boltzmann_calculator.jl
jaemolihm: Removed in 711b5fd (part of the plan-reference comment sweep).

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: Done in 711b5fd. Renamed StateSelection to FilteredStates repo-wide (still a subtype of AbstractBandStates).

src/calculator/run_eph_over_q_and_k.jl
jaemolihm: Done in 711b5fd. _setup_electron_k now returns sel_k (both sibling drivers updated).

src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -129,32 +136,60 @@ function _setup_eph_over_k_and_kq(
...
+            kqpts_irr, ik_to_ikirr_isym_kq = nothing, nothing
         end
-        kqpts_irr, ik_to_ikirr_isym_kq = nothing, nothing
+        # Electron states at k+q (directly, or via IBZ + unfolding for gauge consistency). CPU-only
+        # unfolding — `run_eph_over_k_and_kq` gates el_kq_from_unfolding off for use_gpu.
+        el_kq_save = _setup_electron_kq(model, kqpts, kqpts_irr, ik_to_ikirr_isym_kq,
+            symmetry, el_kq_from_unfolding, window_kq; fourier_mode, use_gpu)
+        # Build the k+q selection from the computed states' per-k band extent (uniform per-state
+        # weights) so the calculator consumes a selection on both paths.
+        sel_kq = _selection_from_computed_states(kqpts, el_kq_save, nelec_kq; nw)
```
jaemolihm: Can we reorganize this mess? Think very hard. Think in terms of what are the branches, and can we group them

src/band_states.jl
```diff
@@ -181,22 +322,73 @@ function find_unfolding_indices(el_i::BandStates, el_f::BandStates, symmetry)
...
+       weight = isempty(s.weights) ? s.kpts.weights[s.iks[i]] : s.weights[i])
 Base.iterate(s::BandStates, i::Int = 1) = i > s.n ? nothing : (s[i], i + 1)
 Base.eltype(::Type{<:BandStates{T}}) where {T} =
     NamedTuple{(:ik, :iband, :xk, :e, :weight), Tuple{Int, Int, Vec3{T}, T, T}}
 
-"Per-state BZ weights, gathered from `kpts.weights` (dense length-`n` array)."
-state_weights(s::BandStates) = s.kpts.weights[s.iks]
+"""Per-state BZ weights (dense length-`n` array): the stored per-state `weights` when non-empty
+(e.g. the multigrid double-grid partition), else gathered from `kpts.weights[iks]`."""
+state_weights(s::AbstractBandStates) = isempty(s.weights) ? s.kpts.weights[s.iks] : s.weights
```
jaemolihm: Can we just add s.weights field on BandStates ?

src/filter.jl
```diff
@@ -195,33 +150,148 @@ end
...
+`shift` applies only to the `NTuple` grid path (ignored for a prebuilt `Kpoints`/`GridKpoints`) and
+is mutually exclusive with `symmetry`. Under `mpi_comm` the k-points are split across ranks: the grid
+is built distributed, each rank filters its slice once (single eigensolve pass), then the kept
+k-points and per-k band ranges are redistributed together through the same gather/scatter so they
+stay aligned, and the local below-window counts are summed.
+"""
+function filter_electron_states(kpts_input, nw::Integer, el_ham, window;
+        symmetry=nothing, fourier_mode="gridopt", use_gpu=false, mpi_comm=nothing, shift=(0, 0, 0))
+    if mpi_comm === nothing
+        gkpts, bmin, bmax, nelec = _filter_with_band_ranges(kpts_input, nw, el_ham, window;
```
jaemolihm: bmin, bmax -> ibmin, ibmax

src/filter.jl
```diff
@@ -195,33 +150,148 @@ end
...
+            symmetry, fourier_mode, use_gpu, shift)
+    else
+        kpts_input isa NTuple{3,Integer} ||
+            throw(ArgumentError("filter_electron_states with mpi_comm requires an NTuple grid spec"))
+        (symmetry === nothing || all(shift .== 0)) ||
+            error("nonzero shift and symmetry incompatible (not implemented)")
+        # Build the grid distributed and filter each rank's slice once (single eigensolve pass).
+        kpoints = kpoints_grid(kpts_input, mpi_comm; shift, symmetry)
+        gkpts_l, bmin_l, bmax_l, nelec_l = _filter_with_band_ranges(kpoints, nw, el_ham, window;
```
jaemolihm: ibmin_l, ibmax_l

src/filter.jl
```diff
@@ -195,33 +150,148 @@ end
...
+            fourier_mode, use_gpu)
+        # Redistribute the kept k-vectors, weights, and per-k band ranges through the SAME
+        # gather/scatter — equal-length arrays share one rank-concatenation order and even split, so
+        # the four stay aligned. Sum the local below-window counts.
+        vectors = mpi_scatter(mpi_gather(gkpts_l.vectors, mpi_comm), mpi_comm)
+        weights = mpi_scatter(mpi_gather(gkpts_l.weights, mpi_comm), mpi_comm)
+        bmin    = mpi_scatter(mpi_gather(bmin_l, mpi_comm), mpi_comm)
+        bmax    = mpi_scatter(mpi_gather(bmax_l, mpi_comm), mpi_comm)
+        gkpts   = GridKpoints(Kpoints(length(vectors), vectors, weights, kpts_input))
```
jaemolihm: Don't we have some MPI wrapper for Kpoints ?

src/filter.jl
```diff
@@ -195,33 +150,148 @@ end
...
+        bmax[ik] >= bmin[ik] || continue
+        for b in bmin[ik]:bmax[ik]
+            push!(iks, ik); push!(ibands, b)
+        end
+    end
+    FilteredStates(gkpts, iks, ibands; nw, nstates_base=nelec)
+end
+
+# Convenience: pull nw/el_ham off a Model.
+filter_electron_states(model::Model, kpts_input, window; kwargs...) =
```
jaemolihm: Bit confising kpts_input is 1st or 2nd argument between Model and nw/el_ham case.

src/run_electron_phonon.jl
```diff
@@ -53,8 +53,10 @@ function run_eph_outer_loop_q(
     end
 
     # Generate k points
-    kpoints, iband_min, iband_max, nelec_below_window = filter_kpoints(k_input, nw,
-        model.el_ham, window, mpi_comm_k; fourier_mode)
+    sel = filter_electron_states(k_input, nw, model.el_ham, window; mpi_comm=mpi_comm_k, fourier_mode)
+    kpoints = sel.kpts
+    _br = band_range(sel); iband_min, iband_max = first(_br), last(_br)
```
jaemolihm, line 58: extrema(band_range(sel)) ?

src/run_transport.jl
```diff
@@ -78,15 +78,18 @@ using HDF5
         # Generate k points
         mpi_isroot() && println("Setting k-point grid")
         symmetry_k = use_irr_k ? symmetry : nothing
-        @time kpts, iband_min_k, iband_max_k, nstates_base_k = filter_kpoints(k_input, nw,
-            model.el_ham, window_k, mpi_comm_k, symmetry=symmetry_k; fourier_mode)
+        @time sel_k = filter_electron_states(k_input, nw, model.el_ham, window_k;
+            symmetry=symmetry_k, mpi_comm=mpi_comm_k, fourier_mode)
+        kpts = sel_k.kpts; nstates_base_k = sel_k.nstates_base
```
jaemolihm: one line per expression

src/run_transport.jl
```diff
@@ -78,15 +78,18 @@ using HDF5
...
 
         # Generate k+q points
         mpi_isroot() && println("Setting k+q-point grid")
         # If k_input is a GridKpoint, set shift for kqpts so that qpts includes the Gamma point.
         shift_k = k_input isa GridKpoints ? Vec3{FT}(k_input.shift) : Vec3{FT}(0, 0, 0)
         shift_kq = shift_k .+ shift_q ./ qgrid
-        @time kqpts, iband_min_kq, iband_max_kq, nstates_base_kq = filter_kpoints(qgrid, nw, model.el_ham, window_kq, mpi_comm_k, shift=shift_kq; fourier_mode)
+        @time sel_kq = filter_electron_states(qgrid, nw, model.el_ham, window_kq;
+            shift=shift_kq, mpi_comm=mpi_comm_k, fourier_mode)
+        kqpts = sel_kq.kpts; nstates_base_kq = sel_kq.nstates_base
```
jaemolihm: one line per expression

src/band_states.jl
```diff
@@ -42,8 +67,32 @@ struct BandStates{T, KT <: AbstractKpoints{T}}
...
+
+"""
+    FilteredStates{T, KT<:AbstractKpoints{T}} <: AbstractBandStates{T, KT}
+
+Lean selection of `(k-point, band)` states over a shared `kpts`, before eigenvectors/velocities are
+computed. Mirrors `BandStates` minus `es`/`vs`. Emitted by the k-point/band generators and consumed
+by `compute_electron_states(model, sel, …)`, which computes eigenvectors/velocities for exactly the
+per-k `band_extent` bands. `electron_states_to_BandStates(el_states, sel)` then attaches `es`/`vs`.
+"""
+struct FilteredStates{T, KT <: AbstractKpoints{T}} <: AbstractBandStates{T, KT}
```
jaemolihm, line 85: Are you satisfied about the struct design? Do you see any more fields that will be helpful? (Callsites commonly using) ?

src/run_electron_phonon.jl
jaemolihm: Considered, but reverted: extrema(band_range(sel)) throws on an all-filtered (empty) selection where band_range is 1:0, whereas the first/last form returns (1,0). Kept the behavior-preserving first/last form (193fe63).

src/filter.jl
jaemolihm: Done in 193fe63 (bmin/bmax renamed ibmin/ibmax).

src/run_transport.jl
jaemolihm: Done in 193fe63 (split into one line per expression).

src/band_states.jl
jaemolihm: Done in 193fe63. weights is now always materialized at construction (from kpts.weights[iks] when not given explicitly), so state_weights and getindex return the field directly; the empty-sentinel derive branch is gone.

src/filter.jl
jaemolihm: Done in 193fe63. The MPI branch reuses mpi_gather_and_scatter(gkpts_l) for the k-points; the per-k band ranges ride along through a parallel gather/scatter with the same concat + even-split so they stay aligned, and the hand-rolled Kpoints/GridKpoints rebuild is gone. Verified byte-identical to serial on 2- and 3-rank mpiexec runs.

src/band_states.jl
jaemolihm: Done in 193fe63: kept the field set as-is. Every field has a distinct O(1) consumer (n for length; nband/nband_ignore for O(1) band_range and the state_index row offset; kpts/iks/ibands are the selection; weights the per-state BZ weight; nstates_base the auto-mu count; indmap the O(1) (ik,iband)-to-state lookup; band_extent the per-k range for compute_electron_states and the GPU k-side projection; es/vs the intrinsic energy/velocity). The one plausible new field (a per-k state-offset prefix to make ind_range_for_k_range O(1)) is a documented non-issue today, so not added speculatively.

src/filter.jl
jaemolihm: Fixed in 193fe63. The Model convenience is reordered kpts-first: filter_electron_states(kpts_input, model, window; ...), consistent with the core filter_electron_states(kpts_input, nw, el_ham, window; ...). No dispatch ambiguity (3-arg Model vs 4-arg Integer).

src/run_transport.jl
jaemolihm: Done in 193fe63 (split into one line per expression).

src/filter.jl
jaemolihm: Done in 193fe63 (bmin_l/bmax_l renamed ibmin_l/ibmax_l).

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: Done in 193fe63. The inlined k / k+q setup is gone: run_eph_over_k_and_kq now calls _setup_electron_k (extended with a prebuilt-FilteredStates pass-through, the branch it was the only driver not to reuse) plus a new _setup_selection_kq helper. Branches named: prebuilt-selection vs grid; (grid only) symmetry IBZ-filter+unfold vs direct; and the gauge-consistent el_kq_from_unfolding path, which stays special inside _setup_electron_kq. About 40 fewer lines, byte-identical.

src/calculator/eph_setup_common.jl
```diff
@@ -6,22 +6,32 @@
...
+    br = band_range(sel_k)
+    iband_min, iband_max = first(br), last(br)
+    # Compute states for exactly the selection's per-k band extent (identical to the window path,
+    # which applies the same per-k in-window range via set_window!, but reuses the selection directly).
     el_k_save = maybe_time(verbosity) do
-        compute_electron_states(model, kpts, el_k_quantities, window_k; fourier_mode, use_gpu)
+        compute_electron_states(model, sel_k, el_k_quantities; fourier_mode, use_gpu)
     end
-    (; kpts, iband_min, iband_max, nelec_below_window_k, el_k_save)
+    # Surface `sel_k` so the driver can forward it to the calculator fan-out: calculators that solve μ
```
jaemolihm: Delete this comment

src/calculator/eph_setup_common.jl
```diff
@@ -6,22 +6,32 @@
...
-        filter_kpoints(kpts_input, nw, model.el_ham, window_k, mpi_comm_k; symmetry, fourier_mode, use_gpu)
-    end
+    sel_k = kpts_input isa FilteredStates ? kpts_input :
+        maybe_time(verbosity) do
+            filter_electron_states(kpts_input, nw, model.el_ham, window_k; symmetry, fourier_mode, use_gpu, mpi_comm=mpi_comm_k)
+        end
+    kpts = sel_k.kpts
+    br = band_range(sel_k)
+    iband_min, iband_max = first(br), last(br)
+    # Compute states for exactly the selection's per-k band extent (identical to the window path,
```
jaemolihm: Delete this comment, do empty line

src/calculator/eph_setup_common.jl
```diff
@@ -47,6 +57,56 @@ function _setup_electron_kq(
...
+# states, returning `(; sel_kq, kqpts, el_kq_save)`. Three sub-branches:
+#   (1) a prebuilt full-BZ `FilteredStates` (e.g. from `unfold_band_states`) — consumed as-is;
+#   (2) a grid — filtered to the window (IBZ-reduced then `unfold_kpoints` to the full BZ under
+#       symmetry, else filtered directly), yielding `kqpts` + `nelec_kq`;
+#   (3) the electron states — computed directly, or via the gauge-consistent IBZ→full unfolding
+#       inside `_setup_electron_kq` (the one path that must stay special, so `el_f` is the exact
+#       symmetry unfolding of `el_i` for the interpolate=false δf feedback).
+# The grid path wraps the computed states into a `FilteredStates` via `_selection_from_computed_states`
+# so the calculator sees a selection on both paths.
+function _setup_selection_kq(model, kqpts_input;
```
jaemolihm: _setup_electron_kq

src/calculator/eph_setup_common.jl
```diff
@@ -47,6 +57,56 @@ function _setup_electron_kq(
...
+        window_kq, mpi_comm_q, symmetry, el_kq_from_unfolding, fourier_mode, use_gpu = false, verbosity = 1)
+    (; nw) = model
+    quantities = ["eigenvalue", "eigenvector", "velocity", "position"]
```
jaemolihm: this should be passed from callsite

src/calculator/eph_setup_common.jl
```diff
@@ -62,13 +122,13 @@ end
 # (backend / verbosity) forward through `kwargs`.
 function _setup_calculators!(
         calculators, kpts, qpts, el_k_save;
-        nw, nmodes, rng_band, el_states_kq, kqpts,
-        nelec_below_window_k, nelec_below_window_kq, nchunks_threads, kwargs...,
+        nw, nmodes, rng_band, el_states_kq, kqpts, nchunks_threads,
+        sel_k = nothing, sel_kq = nothing, kwargs...,
```
jaemolihm: No default for sel_k, sel_kq

src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -130,30 +137,16 @@ function _setup_eph_over_k_and_kq(
...
-            filter_kpoints(kqpts_input, nw, model.el_ham, window_kq, mpi_comm_q; fourier_mode, use_gpu)
-        end
-        kqpts_irr, ik_to_ikirr_isym_kq = nothing, nothing
-    end
-
-    # Electron states at k+q (directly, or via IBZ + unfolding for gauge consistency). CPU-only
-    # unfolding — `run_eph_over_k_and_kq` gates el_kq_from_unfolding off for use_gpu.
-    el_kq_save = _setup_electron_kq(model, kqpts, kqpts_irr, ik_to_ikirr_isym_kq,
-        symmetry, el_kq_from_unfolding, window_kq; fourier_mode, use_gpu)
+    (; sel_kq, kqpts, el_kq_save) = _setup_selection_kq(model, kqpts_input;
```
jaemolihm: (; kqpts, el_kq_save, sel_kq)

src/band_states.jl
```diff
@@ -54,27 +103,72 @@ function _build_indmap(n, nk, nband, nband_ignore, ik, iband)
...
-(per-state velocity) defaults to empty.
+(per-state velocity) defaults to empty. `weights` (per-state BZ weight) is always stored
+length-`n`: when the caller passes an empty `weights`, it is materialized from `kpts.weights[ik]`.
+`band_extent` defaults to the per-k span of `ik`/`iband`.
 """
 function BandStates(kpts::AbstractKpoints{T}, ik::AbstractVector{<:Integer},
         iband::AbstractVector{<:Integer}, e::AbstractVector;
-        nw::Integer, v::AbstractVector = Vec3{T}[], nstates_base = zero(T)) where {T}
+        nw::Integer, v::AbstractVector = Vec3{T}[], weights::AbstractVector = T[],
+        nstates_base = zero(T), band_extent = nothing) where {T}
```
jaemolihm: is non-default band_extent ever useful?

src/band_states.jl
```diff
@@ -54,27 +103,72 @@ function _build_indmap(n, nk, nband, nband_ignore, ik, iband)
...
+end
+
+"""
+    FilteredStates(kpts, ik, iband; nw, weights, nstates_base, band_extent)
+
+Build a `FilteredStates` over the shared grid `kpts` from the selected `(ik[i], iband[i])` pairs.
+`weights` (per-state BZ weight) is always stored length-`n`: an empty `weights` is materialized from
+`kpts.weights[ik]`. `band_extent` (per-k band range) defaults to the per-k span of `ik`/`iband`.
+"""
+function FilteredStates(kpts::AbstractKpoints{T}, ik::AbstractVector{<:Integer},
```
jaemolihm: iks, ibands

Is weights in filtered states ever useful?

src/band_states.jl
```diff
@@ -122,6 +216,54 @@ function electron_states_to_BandStates(el_states::Vector{ElectronState{T}},
...
 
+"""
+    _selection_from_computed_states(kpts, el_states, nstates_base; nw) -> FilteredStates
+
+Build a `FilteredStates` from already-computed `ElectronState`s: each state's per-k band extent is
+`el.rng`, per-state weights are left empty (uniform ⇒ derived from `kpts.weights`). Used by the
+driver's sugar path to wrap the filter+compute result into a selection, so the calculator consumes a
+selection on both the sugar and prebuilt-selection paths. Promotes `kpts` to `GridKpoints`.
+"""
+function _selection_from_computed_states(kpts, el_states, nstates_base; nw)
```
jaemolihm: electron_states_to_FilteredStates

src/filter.jl
```diff
@@ -195,33 +150,148 @@ end
...
-            push!(vectors, xk)
-            push!(weights, kpts2.weights[i])
+    # Shared grid points 1..kpts1.n are the fine points; coarse-only points are appended. Per shared
+    # point collect (band => weight): narrow bands at the fine weight, then wide-only bands at the
+    # coarse weight (dedup: a band already present from the narrow set keeps the fine weight).
+    pt_vectors = collect(kpts1.vectors)
+    pt_weight  = collect(kpts1.weights)
+    bandw = [Dict{Int, T}() for _ in 1:kpts1.n]
+    for i1 in 1:kpts1.n, b in bmin1[i1]:bmax1[i1]
+        bandw[i1][b] = kpts1.weights[i1]
```
jaemolihm: bandw -> band_weights

src/filter.jl
```diff
@@ -195,33 +150,148 @@ end
...
+wide-only bands at the coarse weight while the fine point's narrow bands stay at the fine weight — a
+single physical node carrying two weights across its bands, which a merged single-window k-set
+cannot express. `nstates_base` is the coarse full-grid below-`window2` carrier count. The shared
+grid is stamped with the FINE `ngrid` (`nks1`) so every k lies on a common grid for the q-lookup;
+its per-point `weights` are informational (the per-state `weights` on the selection are the
+authoritative BZ weights).
 """
-function filter_kpoints_multigrid(nks1, nks2, window1, window2, nw, el_ham;
-                                  fourier_mode="gridopt", symmetry=nothing)
+function filter_electron_states_multigrid(nks1, nks2, window1, window2, nw, el_ham;
```
jaemolihm: is there a possibility where state (ib, ik) has weight other than kpts.weights[ik] ?

src/filter.jl
```diff
@@ -195,33 +150,148 @@ end
...
-    (; vectors, weights, ngrid) = kpts1
-    for i in 1:kpts2.n
-        xk = kpts2.vectors[i]
-        if xk_to_ik(xk, kpts1) === nothing
-            push!(vectors, xk)
-            push!(weights, kpts2.weights[i])
+    # Shared grid points 1..kpts1.n are the fine points; coarse-only points are appended. Per shared
+    # point collect (band => weight): narrow bands at the fine weight, then wide-only bands at the
+    # coarse weight (dedup: a band already present from the narrow set keeps the fine weight).
+    pt_vectors = collect(kpts1.vectors)
```
jaemolihm: xks

src/filter.jl
```diff
@@ -195,33 +150,148 @@ end
...
+    pt_weight  = collect(kpts1.weights)
```
jaemolihm: k_weights

src/filter.jl
```diff
@@ -195,33 +150,148 @@ end
...
+function filter_electron_states_multigrid(nks1, nks2, window1, window2, nw, el_ham;
+                                  fourier_mode="gridopt", symmetry=nothing, use_gpu=false)
 
     all(mod.(nks1, nks2) .== 0) || throw(ArgumentError("nks1 must be a multiple of nks2"))
 
-    kpts1, = filter_kpoints(nks1, nw, el_ham, window1; symmetry, fourier_mode)
-    kpts1 = GridKpoints(kpts1)
-    kpts2, = filter_kpoints(nks2, nw, el_ham, window2; symmetry, fourier_mode)
-    kpts2 = GridKpoints(kpts2)
+    kpts1, bmin1, bmax1, _            = _filter_with_band_ranges(nks1, nw, el_ham, window1; symmetry, fourier_mode, use_gpu)  # fine, narrow
```
jaemolihm: bmin/bmax -> ibmin / ibmax

src/filter.jl
```diff
@@ -195,33 +150,148 @@ end
...
+single physical node carrying two weights across its bands, which a merged single-window k-set
+cannot express. `nstates_base` is the coarse full-grid below-`window2` carrier count. The shared
+grid is stamped with the FINE `ngrid` (`nks1`) so every k lies on a common grid for the q-lookup;
+its per-point `weights` are informational (the per-state `weights` on the selection are the
+authoritative BZ weights).
 """
-function filter_kpoints_multigrid(nks1, nks2, window1, window2, nw, el_ham;
-                                  fourier_mode="gridopt", symmetry=nothing)
+function filter_electron_states_multigrid(nks1, nks2, window1, window2, nw, el_ham;
+                                  fourier_mode="gridopt", symmetry=nothing, use_gpu=false)
```
jaemolihm, line 259: Assert nks1 is multiple of nks2, and window1 is smaller than window2

src/calculator/eph_setup_common.jl
jaemolihm: Done in 15f3458. Deleted the comment.

src/calculator/eph_setup_common.jl
jaemolihm: Done in 15f3458. Deleted the comment and left a blank line before the compute step.

src/calculator/eph_setup_common.jl
jaemolihm: Done in 15f3458. Promoted this helper to `_setup_electron_kq` (parallel to the k-side `_setup_electron_k`); the inner electron-states-only helper it calls was renamed `_compute_electron_states_kq` to free the name.

src/calculator/eph_setup_common.jl
jaemolihm: Done in 15f3458. The k+q quantities are now passed in via `el_kq_quantities` (supplied by the driver callsite) rather than hardcoded in the shared helper; threaded through to `_compute_electron_states_kq` as `quantities`.

src/calculator/eph_setup_common.jl
jaemolihm: Done in 15f3458. `sel_k`/`sel_kq` are now required kwargs (no `nothing` defaults); the two sibling drivers forward their `sel_kq` explicitly (`nothing` when k+q is not precomputed).

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: Done in 15f3458. Return reordered to `(; kqpts, el_kq_save, sel_kq)` and the callsite destructure matches.

src/band_states.jl
jaemolihm: Done in 15f3458. It was never passed non-default by any caller, so I dropped the `band_extent` constructor kwarg from both `BandStates` and `FilteredStates`; it is always computed from the per-k span of `iks`/`ibands`.

src/band_states.jl
jaemolihm: Done in 15f3458. Renamed the ctor args to `iks`/`ibands`. On the weights question: yes -- the multigrid generator (`filter_electron_states_multigrid`) assigns per-(k,band) weights that differ from `kpts.weights[ik]` (fine bands at the fine weight, wide-only bands at the coarse weight), and `unfold_band_states` divides by star size, so the per-state `weights` field is load-bearing and cannot be derived from `kpts.weights`.

src/band_states.jl
jaemolihm: Done in 15f3458. Renamed `_selection_from_computed_states` -> `electron_states_to_FilteredStates` (exported, peer of `electron_states_to_BandStates`).

src/filter.jl
jaemolihm: Done in 15f3458. Renamed `bandw` -> `band_weights`.

src/filter.jl
jaemolihm: Done in 15f3458. Renamed `pt_vectors` -> `xks`.

src/filter.jl
jaemolihm: Done in 15f3458. Renamed `pt_weight` -> `k_weights`.

src/filter.jl
jaemolihm: Done in 15f3458. Renamed `bmin1/bmax1/bmin2/bmax2` -> `ibmin1/ibmax1/ibmin2/ibmax2`.

src/filter.jl
jaemolihm: Done in 15f3458. The `nks1` multiple-of-`nks2` assert was already present; added an assert that the fine `window1` is contained in the coarse `window2`.

src/filter.jl
jaemolihm: Yes -- that is exactly what this multigrid generator produces. A state `(ib, ik)` on the fine grid inside the narrow window gets the fine BZ weight; a coarse-only (wide-window, not narrow) band at the same physical node gets the coarse weight. So a single shared node carries two different weights across its bands, none of which is a plain `kpts.weights[ik]`. That is why `FilteredStates` stores per-state `weights` rather than deriving from `kpts.weights`. (Leaving this open in case you want to discuss the design further.)

jaemolihm: I see, you are right. Could you document this? Just say if at a given fine grid k point ik, band ib1 is inside the narrow window, and ib2 is outside the narrow window but inside the wide window, then (ik, ib1) should have fine-grid weight, but (ik, ib2) should have coarse grid weight.

src/filter.jl
```diff
@@ -195,33 +150,150 @@ end
...
+        else
+            for b in Wb
+                haskey(band_weights[i1], b) && continue   # narrow band already at the fine weight
+                band_weights[i1][b] = kpts2.weights[i2]
+            end
         end
     end
 
-    GridKpoints(Kpoints(length(vectors), vectors, weights, ngrid))
+    iks = Int[]; ibands = Int[]; wstate = T[]
```
jaemolihm: wstate -> weights

src/filter.jl
```diff
@@ -195,33 +150,156 @@ end
...
+    # coincides with, record that wide band range and coarse weight (`wide_lo`/`wide_hi`/`wide_w`, 0 =
+    # no coincident coarse node); coarse nodes with no surviving fine point become appended points
+    # (`extra_*`). No dedup or per-band sort is needed since the ranges are contiguous and nested.
+    xks = collect(kpts1.vectors)
+    k_weights = collect(kpts1.weights)
+    wide_lo = zeros(Int, kpts1.n); wide_hi = zeros(Int, kpts1.n); wide_w = zeros(T, kpts1.n)
+    extra_lo = Int[]; extra_hi = Int[]; extra_w = T[]
+    for i2 in 1:kpts2.n
+        i1 = xk_to_ik(kpts2.vectors[i2], kpts1)   # exact: a coarse node lies on the fine grid
+        if i1 === nothing
```
jaemolihm: Add comment # coarse k point i2 not part of fine grid

src/filter.jl
```diff
@@ -195,33 +150,156 @@ end
...
-    kpts2 = GridKpoints(kpts2)
+# Convenience: pull nw/el_ham off a Model. Kept kpts-first to match the core signature; dispatches on
+# the 2nd argument being a `Model` (no ambiguity with the `(kpts_input, nw::Integer, el_ham, window)`
+# core method).
+filter_electron_states(kpts_input, model::Model, window; kwargs...) =
+    filter_electron_states(kpts_input, model.nw, model.el_ham, window; kwargs...)
 
-    # Merge kpts1 and kpts2, remove duplicates
+"""
+    filter_electron_states_multigrid(nks1, nks2, window1, window2, nw, el_ham;
```
jaemolihm: nks1, nks2 -> nks_f, nks_c
same for window1, window2

src/filter.jl
jaemolihm: Done in 45d5f0e. Documented in the docstring: at a fine-grid k point ik, a band ib1 inside the narrow window gets the fine-grid weight, while a band ib2 outside the narrow window but inside the wide window gets the coarse-grid weight. (Also restructured the assembly per your other comment: merge the grids, loop fine points for narrow bands, loop coarse points for wide-only bands, then sort by (ik, band) to keep each k's states contiguous for the GPU tiling.)

src/filter.jl
jaemolihm: Done in 45d5f0e. Renamed the per-state weight array `wstate` -> `weights`.

src/filter.jl
jaemolihm: Done in 45d5f0e. Added the `# coarse k point i2 not part of fine grid` comment in that branch (now inside the grid-merge loop).

src/filter.jl
jaemolihm: Done in 45d5f0e. Renamed `nks1`/`nks2` -> `nks_f`/`nks_c` and `window1`/`window2` -> `window_f`/`window_c` throughout the function + docstring (and the derived locals `kpts1`/`ibmin1`... -> `kpts_f`/`ibmin_f`... for consistency). Callers pass positionally, so no callsite change.

PR #11, opened 2026-08-02
Speed up the GPU outer-k e-ph loop 3.3x: dense q-index lookup and k+q-convention phase hoist

jaemolihm:
## Motivation

The GPU outer-k e-ph loop (`_loop_eph_over_k_and_kq_gpu`) ran at **43% kernel-time utilization** on a realistic BTE run — the GPU idled while one host thread did per-pair index work. Two changes fix that, and both apply to every model. A third (a k-strip GEMM) was built, measured, and then **removed**: it was worth 1.23× but only for single-atom cells at a narrow window, which is not a class worth a permanent loop level.

Cu, nk=150, ±0.2 eV, symmetry, A100-SXM4-80GB (`workergpu053`, SW-capped 225 W), 16 threads:

| | loop wall | kernel-time util |
|---|---|---|
| `main` (`57a42b1`) | 101.44 s | 43% |
| + dense q-index (1.70×) | 59.7 s | |
| + k+q-convention phase hoist (1.89×) | 31.6 s | |
| **this branch** (`8e8b3cc`) | **30.89 s** | ~93% |

**3.28× end to end.** At nk=200 the loop is 177.1 s. `combine_kpoint_grids` in setup also went **76.86 → 2.69 s (28.5×)** at nk=200 (#12, merged into this branch).

## Change

**Dense q-index** (`src/common/kpoints.jl`). The per-pair `iq` lookup probed a 142.6 MB `Dict` for all nk×nkq = 1.11e9 pairs, plus a dependent random gather — both past L3. Replaced by a size-gated dense `Array{Int,3}` of dims `(ng3,ng2,ng1)` indexed `[c3+1,c2+1,c1+1]`, whose linear index is exactly `hash+1`. All lookups route through one `_ik_from_hash`, so the CPU e-ph / transport / screening paths get it free.

- The `qpts_is_full ? hash+1` fast path was **dead code and always had been** (`shift_center!` folds coords signed, `_hash_xk` applies `mod`) — verified false even on genuinely full 2³/3³/4³/6³ grids. Every run of this loop had always paid the `Dict`.
- `index = :auto/:dense/:dict` override on `GridKpoints` for testing the fallback.

**k+q-convention phase hoist** (`src/wannier_to_bloch_batched.jl`, `run_eph_over_k_and_kq.jl`). Uses
`exp(2πi R_p·q) = exp(2πi R_p·x_{k+q}) · conj(exp(2πi R_p·x_k))` (verified to 1.1e-14 including mod-G reduction) to store `g(k, R_p)` in the k+q convention, which makes the kR→kq phase **k-independent**:

- before: rebuild the `(nr_ep × nq)` phase per (k, q-tile) — 4918 × ntiles times
- after: build it once per q-tile, reuse for every k in the batch — 20 times

Loop reorders to **k-batch → q-tile → k** (+ q on device). Deletes the child `WannierObject`, the second interpolator, the per-k `update_op_r!`, `qs_batch` (a 66 MB dependent gather per pair), `ikqs_host`/`ikqs_dev`, and `rdotk` repo-wide.

**Contract break.** The batched outer-k loop no longer fires the per-k `OuterIteration` bracket — under the reorder a k's work only completes at the end of the k-batch, so the hook cannot mean what its name says. `BatchedMode` outer-k has `OuterIterationBatch` only. Pinned in `test_calculator_contract.jl`, documented in `docs/writing_a_calculator.md` and the README table. Both in-tree calculators already treated it as a no-op.

**`EPDataQBatched.ikqs` is a `UnitRange`** — `isbits`, so it rides in the kernel launch parameters instead of costing a device buffer and one global load per scatter thread.

## Measured, then removed

Two constructs were built, measured, and reverted. The numbers are kept so nobody redoes the experiments (full records in `GPU_PROGRESS.md`).

**k-strip kR→kq GEMM** — reverted in `8e8b3cc`, **−250 lines**. Stacked `nk_b` outer k into one tall GEMM. Real and re-measured at tip (Cu nk=200, shared setup, same binary via a `Ref` override, alternating arms):

| arm | reps (s) | median |
|---|---|---|
| `nk_b = 7` | 145.35 / 144.86 / 144.92 | 144.92 |
| `nk_b = 1` | 178.40 / 178.31 / 178.10 | 178.31 |

1.230×, 33.4 s, **18.7% of loop wall**. Removed anyway, because `_krkq_strip_width` has two gates and neither is escapable in practice:

| nw | nmodes | in-window bands | ndata | nk_b |
|---|---|---|---|---|
| 7 | 3 | 1 | 21 | 7 |
| 7 | 3 | 3 | 63 | 3 |
| 7 | 3 | 7 (full band) | 147 | **1** |
| 7 | 6 | 1 | 42 | **1** |
| 32 | 12 | 1 | 384 | **1** |

`nw*nmodes ≤ 24` is **model-only**, so no window choice rescues a multi-atom cell (nmodes=6 needs nw≤4; nmodes=12 needs nw≤2), and more in-window bands closes the `ndata < 128` gate independently. It engaged only for Cu at ±0.2 eV and Na. Removal also cut per-q device memory 5.1 → 3.6 kB/q, so q-tiles get bigger on exactly the large-`nw` models the strip never helped.

**Per-strip `iq` staging** — reverted in `f48b597`, −47 lines. Justified on "7× fewer stream drains", but a `:nocopy` probe measured that mechanism at **−0.64 s** (the blocking `copyto!` usefully throttles host look-ahead), and a `:null` arm replacing the build with a constant fill measured level with the shipped code — the host build is already free. Measured worth: 0.02% against 0.16–0.42 s of session drift.

## Test

`Sᵢ` is the correctness target throughout. `Sₒ` is **not** run-to-run reproducible (`CUDA.@atomic` fold, ~1e-14 relative), and `Base.hash` on a large array **samples elements**, so bitwise claims below use a full-array FNV fold.

- **`Sᵢ` bitwise identical** across: dense-vs-`Dict` index, 1-tile vs 4-tile emission order, both strip widths, and strip-present vs strip-removed (FNV over all 1.005e8 elements, `1224223090643646050` on both sides of `f48b597..8e8b3cc`).
- **Index equivalence** — `_ik_from_hash(kpts,h) == get(dict,h,0)` for **every** `h in 0:prod(ngrid)-1` across seven grid constructions.
- **Interpolation** — band energies, velocities, phonon ω and raw `H(k)`/`D(q)` bitwise identical before/after on a fixed 8³ grid, CPU and GPU. A 256-bit `BigFloat` reference puts both e-ph factorizations at ~1 ulp of `max|g|`.
- **CPU vs GPU** — `rdiff = 8.1e-16` on the in-tree driver comparisons.
- **Suites** — `test_gpu.jl` (incl. exact byte-accounting pins), `test_iq_build.jl`, `test_kpoints.jl`, `test_calculator_contract.jl`, `test_gpu_boltzmann_calculator.jl`: **5004 pass**. MigdalEliashberg 41/41 with GPU `EliashbergCalculator` 9/9 and GPU-vs-CPU `solve!` 4/4 *actually running* — CUDA is in ME's `[extras]`, so those two testsets silently skip unless the env carries both `Lehmann` and `CUDA`.

**Crash this branch shipped and fixed** (`4decfc3`, worth a look): `_krkq_strip_width` returned 1 exactly when `nw*nmodes > 24`, i.e. exactly when there is no fused rotation kernel, so those models took the cuBLAS two-GEMM branch — which reshapes `g`. The operand `view(g_strip5, :, :, :, kloc, rng_q)` *is* dense, but its `reshape` is a `ReshapedArray` with **no device pointer**, so every outer-k GPU run with `nw*nmodes > 24` aborted at the first rotation. Dense is necessary, not sufficient. No test caught it because no in-repo GPU model exceeds the threshold (Cu 7·3, Pb 4·3). Moot now that the strip is gone, but the `_is_dense` guard and the `nw=6, nmodes=6` two-GEMM coverage remain.

Measurement notes for anyone re-running: `nvidia-smi utilization.gpu` marks an interval busy if *any* kernel ran in it and reads 91–94% here, so the table uses kernel time. This host's A100 is SW-capped at 225 W — sustained FP64 drops the SM clock 1275 → 870 MHz, so kernel durations are **not** comparable between runs at different occupancy.

## Known pre-existing, not addressed

- `xk_to_ik` on the `GridKpoints{T}()` placeholder throws `DivideError` (`ngrid == (0,0,0)` → `mod(·, 0)`). Predates this branch, affects both index paths.
- `plan_batch`'s `warn = false` kwarg was orphaned by `4decfc3`; it has a test and is a general opt-out, so it is left in place.
- Companion commit `31933f0` on MigdalEliashberg.jl `gpubte-opt` (comment-only, calculator brackets) — that repo has no git remote, so it is local-only and not reviewable here.

### DIFF
--- a/src/calculator/AbstractCalculator.jl
+++ b/src/calculator/AbstractCalculator.jl
@@ -18,7 +18,8 @@ Users subtype `AbstractCalculator` and implement:
   around one outer iteration (`OuterIteration()`) or one batch of outer iterations
   (`OuterIterationBatch()`). There is NO default: a calculator must define these for every
   (scope, loop-mode) combination its supported loops fire, even as an explicit no-op (`= nothing`);
-  a missing method is a loud error, not a silent skip.
+  a missing method is a loud error, not a silent skip. The GPU outer-k loop fires only
+  `OuterIterationBatch` (see the bracket comment below).
 
 Optionally:
 * `eph_batched_bytes_per_point(calc, PayloadType; nw, nmodes)` — per-point device scratch (bytes)
@@ -168,9 +169,12 @@ struct OuterIterationBatch end   # one batch of consecutive outer iterations (GP
 # (scope, loop-mode) combination the drivers fire on it, even when it wants a no-op — a missing
 # method is a loud error, never a silent skip (a silently-skipped bracket is a hard-to-find bug). The
 # combinations follow what the calculator `supports`: OuterIteration/SingleMode (CPU loops),
-# OuterIteration/BatchedMode (GPU outer-k per-k bracket; GPU outer-q per-q accumulator), and
-# OuterIterationBatch/BatchedMode (GPU outer-k per-batch). OuterIterationBatch/SingleMode is never
-# fired. A calculator that does nothing at a scope defines an explicit no-op (`= nothing`).
+# OuterIteration/BatchedMode (GPU outer-q per-q accumulator), and OuterIterationBatch/BatchedMode
+# (GPU outer-k per-batch). OuterIterationBatch/SingleMode is never fired, and neither is
+# OuterIteration/BatchedMode in the GPU OUTER-K loop: there the q-tile loop runs outside the k loop,
+# so a single k's work is spread over the whole batch and has no per-k begin/end point — a
+# calculator needing a per-k device reduction there must do it at OuterIterationBatch scope.
+# A calculator that does nothing at a scope defines an explicit no-op (`= nothing`).
 function calculator_begin!(calc::AbstractCalculator, scope, ctx)
     error("calculator_begin!($(typeof(calc)), ::$(typeof(scope)), ::$(typeof(ctx))) is not " *
           "defined. Every calculator must define the begin/end brackets for the (scope, loop-mode) " *
--- a/src/calculator/run_eph_over_k_and_kq.jl
+++ b/src/calculator/run_eph_over_k_and_kq.jl
@@ -661,6 +661,10 @@ function _loop_eph_over_k_and_kq_gpu(
     # Phonon eigenvectors `u` and frequencies `e` depend only on iq, so (like ukqs_all_dev above)
     # collect the full q-grid stacks on the host and copy to the device once, then gather per
     # batch on the device by index (below).
+    # TODO: these two are the only fields of `ph_save` this loop ever reads, and
+    # `_compute_phonon_states_gpu!` built them on the device before scattering them into per-q
+    # `PhononState`s — so this gather + H2D undoes a D2H. Have the setup hand `use_gpu` the device
+    # stacks directly and drop this block; see the TODO at `_scatter_phonon_states!`.
     uph_all_dev = similar(epmat_dev.op_r, Complex{FT}, nmodes, nmodes, qpts.n)
     ωq_all_dev  = similar(epmat_dev.op_r, FT, nmodes, qpts.n)
     let uph_all_host = Array{Complex{FT}}(undef, nmodes, nmodes, qpts.n),
--- a/src/calculator/run_eph_over_k_and_kq.jl
+++ b/src/calculator/run_eph_over_k_and_kq.jl
@@ -621,15 +621,22 @@ function _loop_eph_over_k_and_kq_gpu(
               "$(round(per_point / 1e3, digits = 1)) kB/q; q-batch size = $nq_batch_max, " *
               "kR->kq strip nk_b = $nk_b"
     end
-    # The strip's extra `ws.g` rows are spent out of the same per-q budget as the phase tile. At the
-    # shapes where the strip switches on they are a small share of it, but a model with a large
-    # `ndata` and a small `nr_ep` inverts that; say so rather than silently starving the q-tile.
-    strip_bytes = sizeof(Complex{FT}) * ndata_ekpR * (nk_b - 1)
-    if nk_b > 1 && nq_batch_max < nq_batch_cap && strip_bytes * 4 > per_point
-        @warn "GPU outer-k: the kR->kq strip (nk_b = $nk_b) takes " *
-              "$(round(Int, 100 * strip_bytes / per_point))% of the per-q device budget while the " *
-              "q-tile is memory-limited ($nq_batch_max of $nq_batch_cap). Lower _KRKQ_GEMM_M_TARGET if this " *
-              "run is slower than one with nk_b = 1."
+    # The strip widens `ws.g`, which is spent out of the same per-q budget as the phase tile. At the
+    # shapes where the strip switches on it is a small share of that budget, but a model with a large
+    # `ndata` and a small `nr_ep` inverts the ratio. Report the quantity that actually matters — how
+    # much q-tile the strip cost — by re-asking the accountant at `nk_b = 1`, so this stays right if
+    # another buffer ever becomes strip-sized. Only when the tile is memory-bound (`< nq_batch_cap`):
+    # a user-supplied `nq_batch_max` is not the strip's fault.
+    if nk_b > 1 && nq_batch_max < nq_batch_cap && mpi_isroot()
+        per_point_1, _ = _outer_k_staging_bytes(; nw, nbandk_max, nmodes, nr_ep, nk, nkq,
+            nq_grid = qpts.n, nk_batch_max, calculators,
+            ndata_epmat = epmat_dev.ndata, nr_epmat = epmat_dev.nr, nk_b = 1, FT)
+        nq_batch_1 = plan_batch(backend, per_point_1, committed, nq_batch_cap; warn = false)
+        if nq_batch_max * 2 < nq_batch_1
+            @warn "GPU outer-k: the kR->kq strip (nk_b = $nk_b) more than halved the q-tile " *
+                  "($nq_batch_1 -> $nq_batch_max of $nkq k+q). Lower `nk_outer_batch_max` (it " *
+                  "clamps nk_b) if this run is slower than one with nk_b = 1."
+        end
     end
 
     # ----- persistent workspace (allocated once, reused across all (k, q)) -----
@@ -796,6 +803,9 @@ function _loop_eph_over_k_and_kq_gpu(
             # (unit stride down the rows, `lda = ndata_ekpR*nk_batch_max` across) - a plain view,
             # no copy. A partial final strip just uses a shorter row range: `M` is a runtime GEMM
             # dimension, so nothing needs padding. Everything below the GEMM stays strictly per-k.
+            # NOT an invitation to overlap the k of a strip: `g` is the only strip-wide buffer.
+            # `epkq_dev`, `g2_dev`, `uphs_dev`, `ωq_dev` and `kRkq_ws.tmp` are all single-k and are
+            # overwritten by each k below, correct only because the k are serialized on one stream.
             for sstart in 1:nk_b:nk_batch
                 nks  = min(nk_b, nk_batch - sstart + 1)
                 rows = ndata_ekpR * (sstart - 1) + 1 : ndata_ekpR * (sstart - 1 + nks)
--- a/src/calculator/run_eph_over_k_and_kq.jl
+++ b/src/calculator/run_eph_over_k_and_kq.jl
@@ -693,7 +693,7 @@ function _loop_eph_over_k_and_kq_gpu(
     # exact, and `cispi` is exactly symmetric). This is the only consumer of the k coordinates.
     # Out-of-place on purpose: on `CPUBackend` `_kpoints_to_device_matrix` returns a view onto
     # `kpts.vectors`, so negating in place would corrupt the k-points.
-    mxk_dev = -1 .* _kpoints_to_device_matrix(backend, kpts)
+    mxk_dev = _kpoints_to_device_matrix(backend, kpts) .* -1
     xkq_dev = _kpoints_to_device_matrix(backend, kqpts)
 
     # The two Fourier phase matrices of the k+q convention (see `get_eph_RR_to_kR_batched!`):
--- a/src/calculator/run_eph_over_k_and_kq.jl
+++ b/src/calculator/run_eph_over_k_and_kq.jl
@@ -513,6 +513,57 @@ end
 #  small-`nw` regime; a full-band large-`nw` run gets `nk_b = 1` and the loop as it was. k are
 #  visited in the same order and everything below the GEMM is per-k, so the payload stream and the
 #  bracket order are unchanged.
+# ---- per-(k-strip, q-tile) `iq` index build ------------------------------------------------------
+#
+# `prepare_iqs!(builder, iqs_dev, iks, kloc0, qstart, nq)` fills `iqs_dev[1:nq, kloc0 + i - 1]` with
+# the index into `qpts` of `x_{k+q} - x_k` for the `i`-th outer k of `iks` and every k+q of the tile
+# `qstart .+ (0:nq-1)`, leaving the device buffer ready for the phonon gather. The builder owns its
+# host staging matrix and ends with the H2D.
+#
+# The staging is per k-STRIP (`nk_b` columns), not per k: one stream drain per strip instead of
+# `nk_b` of them, for `nk_b * nq_batch_max * 8` B of host and device staging (12.7 MB at Cu nk=150).
+# That drain count is most of what this block costs — see `plans/qindex_build_threaded_vs_device.md`.
+# At `nk_b == 1` it degenerates to per-k staging exactly.
+struct SerialHostIqBuilder{GK}
+    qpts     :: GK
+    xkqs_int :: Matrix{Int}   # (3, nkq), reduced into 0:ngrid[d]-1
+    xks_int  :: Matrix{Int}   # (3, nk),  reduced into 0:ngrid[d]-1
+    iqs_host :: Matrix{Int}   # (nq_batch_max, nk_b) host staging
+end
+
+function prepare_iqs!(b::SerialHostIqBuilder, iqs_dev, iks, kloc0::Int, qstart::Int, nq::Int)
+    _fill_iqs_serial!(b.iqs_host, b.qpts, b.xkqs_int, b.xks_int, iks, kloc0, qstart, nq)
+    # One contiguous 5-arg H2D of the columns just written (a host↔device SubArray copy would fall
+    # back to scalar indexing). Full columns: the rows past `nq` are never read by the gather.
+    copyto!(iqs_dev, size(b.iqs_host, 1) * (kloc0 - 1) + 1,
+            b.iqs_host, size(b.iqs_host, 1) * (kloc0 - 1) + 1, size(b.iqs_host, 1) * length(iks))
+    iqs_dev
+end
+
+# The build itself, as a standalone function so the `iks`/`qstart`/`nq` it closes over are typed
+# arguments rather than `Core.Box`-wrapped captures of the enclosing loop bodies.
+function _fill_iqs_serial!(iqs, qpts, xkqs_int, xks_int, iks, kloc0, qstart, nq)
+    ng1, ng2, ng3 = qpts.ngrid
+    for (i, ik) in enumerate(iks)
+        kloc = kloc0 + i - 1
+        k1, k2, k3 = xks_int[1, ik], xks_int[2, ik], xks_int[3, ik]
+        for j in 1:nq
+            ikq = qstart + j - 1
+            # Integer grid-coord hash for iq (no Float64 normalize/_hash_xk per pair). Both operands
+            # are pre-reduced into 0:ng-1 at setup, so the fold is a compare-and-add.
+            h1 = _wrap_reduced(xkqs_int[1, ikq] - k1, ng1)
+            h2 = _wrap_reduced(xkqs_int[2, ikq] - k2, ng2)
+            h3 = _wrap_reduced(xkqs_int[3, ikq] - k3, ng3)
+            hash = (h1 * ng2 + h2) * ng3 + h3
+            iq = _ik_from_hash(qpts, hash)
+            # 0 = miss on either index.
+            (iq < 1 || iq > qpts.n) && throw(ArgumentError("kq - k = q point not found in precomputed qpts"))
+            iqs[j, kloc] = iq
+        end
+    end
+    iqs
+end
+
 function _loop_eph_over_k_and_kq_gpu(
         model       :: Model{FT},
         kpts, qpts, kqpts,
@@ -708,30 +759,33 @@ function _loop_eph_over_k_and_kq_gpu(
     # convention multiply and keeps a padded (never-read) slice of ep_ekpR_all meaningful.
     fill!(P_k, 1)
 
-    iqs_batch  = Vector{Int}(undef, nq_batch_max)
-    iqs_batch_dev    = similar(epmat_dev.op_r, Int, nq_batch_max)
+    # `iq` index staging, one k-STRIP wide (see `prepare_iqs!`). Filled to 1 so that the rows past a
+    # partial q-tile's width — copied along with the rest, never read by the gather — are a valid
+    # index rather than garbage.
+    iqs_strip     = fill(1, nq_batch_max, nk_b)
+    iqs_strip_dev = similar(epmat_dev.op_r, Int, nq_batch_max, nk_b)
 
     # Integer grid-coord hash for iq, replacing the per-(k,q) float normalize + `_hash_xk`:
-    #   hc_i = mod(xkqs_int[i,ikq] - xks_int[i,ik], ng_i),  hash = (hc1*ng2 + hc2)*ng3 + hc3,
+    #   hc_i = fold(xkqs_int[i,ikq] - xks_int[i,ik], ng_i),  hash = (hc1*ng2 + hc2)*ng3 + hc3,
     # reproducing `_hash_xk` bit-identically with no Float64 in the hot loop. Requires every k and
     # k+q to lie exactly on the q-grid (ngrid a multiple of both meshes) — guaranteed by precompute_ph,
     # asserted above. Only `iq` is needed: the q-VECTOR no longer enters the interpolation (the
     # kR->kq phase is built from x_{k+q}), so this loop gathers phonon data by index only.
-    ng1, ng2, ng3 = qpts.ngrid
+    # Both coordinate lists are reduced into `0:ng-1` here, once, so the per-pair fold is
+    # `_wrap_reduced` (a compare-and-add) instead of `mod` (a runtime integer division). The q-grid
+    # shift is subtracted on the k+q side only: q = x_{k+q} - x_k - shift, and folding it into one
+    # operand keeps it out of the pair loop.
     shq = qpts.shift
     xkqs_int = Matrix{Int}(undef, 3, nkq)
     xks_int  = Matrix{Int}(undef, 3, nk)
-    for ikq in 1:nkq, d in 1:3
-        xkqs_int[d, ikq] = round(Int, (kqpts.vectors[ikq][d] - shq[d]) * qpts.ngrid[d])
+    for ikq in 1:nkq
+        xkqs_int[:, ikq] .= _grid_coords_reduced(kqpts.vectors[ikq], qpts.ngrid, shq)
     end
-    for ik in 1:nk, d in 1:3
-        xks_int[d, ik] = round(Int, kpts.vectors[ik][d] * qpts.ngrid[d])
+    for ik in 1:nk
+        xks_int[:, ik] .= _grid_coords_reduced(kpts.vectors[ik], qpts.ngrid, zero(Vec3{FT}))
     end
 
-    # The payload's k+q index list is the plain range `qstart:qend` of the current q-tile, so hold
-    # 1:nkq on the device once and slice it instead of refilling and re-uploading it per (k, tile).
-    ikqs_all_dev = similar(epmat_dev.op_r, Int, nkq)
-    copyto!(ikqs_all_dev, collect(1:nkq))
+    iq_builder = SerialHostIqBuilder(qpts, xkqs_int, xks_int, iqs_strip)
 
     ωq_dev    = similar(epmat_dev.op_r, FT, nmodes, nq_batch_max)
     g2_dev    = similar(epmat_dev.op_r, FT, nw, nbandk_max, nmodes, nq_batch_max)
@@ -783,9 +837,11 @@ function _loop_eph_over_k_and_kq_gpu(
             # the reason the q-tile loop sits outside the k loop. In the k+q convention it reads
             # x_{k+q} directly, so it is a contiguous slice of the fixed k+q list.
             @views fourier_phase!(P_kq[:, rng_q], irvecp_mat, xkq_dev[:, qstart:qend])
-            # k+q rotations and k+q indices: contiguous slices of prebuilt device stacks (no copy).
+            # k+q rotations: a contiguous slice of the prebuilt device stack (no copy). The payload's
+            # k+q index list is the plain range of this tile — `isbits`, so it rides in the kernel
+            # launch parameters instead of costing a device buffer and a global load per thread.
             ukqs_used = view(ukqs_all_dev, :, :, qstart:qend)
-            ikqs_used = view(ikqs_all_dev, qstart:qend)
+            ikqs_used = qstart:qend
 
             # Strip-mine the k loop for the kR->kq Fourier: one tall GEMM stacks `nks` outer-k
             # into a single left operand, so the k-invariant phase tile is read once per strip
@@ -802,41 +858,31 @@ function _loop_eph_over_k_and_kq_gpu(
                 eph_kR_to_kq_fourier!(view(kRkq_ws.g, 1:ndata_ekpR*nks, rng_q),
                     view(ep_ekpR_2d, rows, :), view(P_kq, :, rng_q))
 
+                # Build this (strip, tile)'s q-index lists; the phonon eigenvectors `u` and
+                # frequencies `e` are gathered on the device by iq below. One entry and one stream
+                # drain for the whole strip.
+                prepare_iqs!(iq_builder, iqs_strip_dev,
+                    iks_batch[sstart:(sstart + nks - 1)], 1, qstart, nq_batch)
+
                 for ik_ind in sstart:(sstart + nks - 1)
                     ik = iks_batch[ik_ind]
-                    # Build this (k, tile)'s q-index list (host-side integers only); the phonon
-                    # eigenvectors `u` and frequencies `e` are gathered on the device by iq below.
-                    for j in 1:nq_batch
-                        ikq = qstart + j - 1
-                        # Integer grid-coord hash for iq (no Float64 normalize/_hash_xk per pair).
-                        h1 = mod(xkqs_int[1, ikq] - xks_int[1, ik], ng1)
-                        h2 = mod(xkqs_int[2, ikq] - xks_int[2, ik], ng2)
-                        h3 = mod(xkqs_int[3, ikq] - xks_int[3, ik], ng3)
-                        hash = (h1 * ng2 + h2) * ng3 + h3
-                        iq = _ik_from_hash(qpts, hash)
-                        # 0 = miss on either index.
-                        (iq < 1 || iq > qpts.n) && throw(ArgumentError("kq - k = q point not found in precomputed qpts"))
-                        iqs_batch[j] = iq
-                    end
+                    kloc = ik_ind - sstart + 1
 
                     # Gather this tile's phonon eigenvectors/frequencies on the device by iq
                     # (uphs_dev[:,:,j] = uph_all_dev[:,:,iq[j]]); ωq gathered here too so the fused kernel
                     # can fold g2 = |ep|²/(2ω) in the same pass. Everything below runs at width nq_batch
-                    # via views into the nq_batch_max-sized buffers, so there is no padded tail.
-                    # H2D copy of just the first nq_batch indices (5-arg contiguous copy — copying a host↔
-                    # device SubArray view instead would fall back to scalar indexing); the device gather
-                    # then reads only `view(iqs_batch_dev, rng_q)`, so the untouched tail is never used.
+                    # via views into the nq_batch_max-sized buffers, so there is no padded tail. This k's
+                    # indices are column `kloc` of the strip staging — contiguous, so no strided read.
                     # `@inbounds`: bounds-checking a device INDEX ARRAY costs a Bool map+reduce kernel and
-                    # a D2H round trip of the result per (k, tile); `iq` is already validated on the host
-                    # (the guard above), so the device-side re-check is pure overhead.
-                    copyto!(iqs_batch_dev, 1, iqs_batch, 1, nq_batch)
-                    @inbounds @views uphs_dev[:, :, rng_q] .= uph_all_dev[:, :, view(iqs_batch_dev, rng_q)]
-                    @inbounds @views ωq_dev[:, rng_q]      .= ωq_all_dev[:, view(iqs_batch_dev, rng_q)]
+                    # a D2H round trip of the result per (k, tile); `iq` is already validated where it is
+                    # built (the guard in `_fill_iqs_serial!`), so the device-side re-check is pure overhead.
+                    iqs_used = view(iqs_strip_dev, rng_q, kloc)
+                    @inbounds @views uphs_dev[:, :, rng_q] .= uph_all_dev[:, :, iqs_used]
+                    @inbounds @views ωq_dev[:, rng_q]      .= ωq_all_dev[:, iqs_used]
 
                     # Gauge rotations for this k, folding g2 = |ep|²/(2ω) into the same fused kernel
                     # pass. This k's Fourier output is the strip's `kloc`-th slice, strided in q when
                     # nk_b > 1 - which only the fused kernel reads, hence the `_krkq_strip_width` gate.
-                    kloc = ik_ind - sstart + 1
                     eph_apply_rotations!(view(epkq_dev, :, :, :, rng_q),
                         _strip_g_slice(kRkq_ws.g, g_strip5, kloc, rng_q), ukqs_used,
                         view(uphs_dev, :, :, rng_q), view(kRkq_ws.tmp, :, :, rng_q);
--- a/src/calculator/run_eph_over_k_and_kq.jl
+++ b/src/calculator/run_eph_over_k_and_kq.jl
@@ -679,8 +679,9 @@ function _loop_eph_over_k_and_kq_gpu(
     irvecp_mat = _irvec_matrix(model.epmat.irvec_next, epmat_dev, FT)
     P_k  = similar(epmat_dev.op_r, Complex{FT}, nr_ep, nk_batch_max)
     P_kq = similar(epmat_dev.op_r, Complex{FT}, nr_ep, nq_batch_max)
-    # A partial k-batch leaves P_k's padded tail columns unwritten; 1 keeps the padded (discarded)
-    # slices of ep_ekpR_all finite.
+    # Defensive: only columns 1:nk_batch are rewritten per batch. `nk_batch_max = min(…, nk)` makes
+    # the first batch full, so the tail is in practice always written, but 1 is the identity of the
+    # convention multiply and keeps a padded (never-read) slice of ep_ekpR_all meaningful.
     fill!(P_k, 1)
 
     iqs_batch  = Vector{Int}(undef, nq_batch_max)
--- a/src/calculator/run_eph_over_k_and_kq.jl
+++ b/src/calculator/run_eph_over_k_and_kq.jl
@@ -478,15 +478,12 @@ end
 #  GPU calculator loop (see README_GPU.md).
 #
 #  This mirrors `_loop_eph_over_k_and_kq` but moves the e-ph Wannier->Bloch interpolation
-#  onto the device using the batched drivers from `wannier_to_bloch_batched.jl`. No CUDA code lives
-#  in the base package: the loop only calls `to_device` and the generic batched drivers, and the
-#  device methods are provided by the CUDA extension. That is not the same as running on any
-#  backend — at `nk_b > 1` the per-k `g` slice handed to `eph_apply_rotations!` is strided in q, and
-#  only the extension's fused kernel reads such an operand (the generic two-GEMM method asserts a
-#  dense `g`). The loop has one call site, inside the `use_gpu` branch, so a CPU backend never
-#  reaches that assert. It is a separate function from the CPU `_loop_eph_over_k_and_kq` because its
-#  control flow — batched over k-batches and q-batches with device staging — differs from the
-#  per-(k,q) CPU loop.
+#  onto the device using the batched drivers from `wannier_to_bloch_batched.jl`. The code here is
+#  backend-generic: it only calls `to_device` and the generic batched drivers, so no CUDA
+#  code lives in the base package (the device methods are provided by the CUDA extension). It
+#  is a separate function from the CPU `_loop_eph_over_k_and_kq` because its control flow —
+#  batched over k-batches and q-batches with device staging — differs from the per-(k,q) CPU loop,
+#  not because it holds any device-specific code.
 #
 #  Supported (all handled upstream in the shared `_setup`, so this loop itself is agnostic to them):
 #    * energy windows (window_k / window_kq): the k side carries only its in-window bands, the
@@ -504,18 +501,12 @@ end
 #  Buffer reuse: device buffers are allocated once before the k loop and reused for every
 #  (k, q), so the loop itself allocates almost nothing.
 #
-#  Loop shape: `k-batch -> q-tile -> k-strip -> k -> q(device)`. The q-tile loop sits OUTSIDE the
-#  per-k loop because the kR->kq Fourier phase is built from the fixed k+q list (the k+q convention,
-#  see `get_eph_RR_to_kR_batched!`) and is therefore the same for every k of the batch: one built
-#  tile is reused `nk_batch` times. Consequence for calculators: a k's work no longer finishes at a
+#  Loop shape: `k-batch -> q-tile -> k -> q(device)`. The q-tile loop sits OUTSIDE the per-k loop
+#  because the kR->kq Fourier phase is built from the fixed k+q list (the k+q convention, see
+#  `get_eph_RR_to_kR_batched!`) and is therefore the same for every k of the batch: one built tile
+#  is reused `nk_batch` times. Consequence for calculators: a k's work no longer finishes at a
 #  single point in the loop, so this loop fires only the `OuterIterationBatch` bracket — there is
 #  no per-k `OuterIteration` bracket in the batched outer-k loop.
-#  The k-strip level runs the kR->kq Fourier for `nk_b` k at once (`_krkq_strip_width`), so that
-#  phase tile is also READ once per strip rather than once per k and the GEMM's M dimension fills a
-#  cuBLAS tile. It helps only where `ndata = nw·nbandk_max·nmodes` is small — a narrow-window,
-#  small-`nw` regime; a full-band large-`nw` run gets `nk_b = 1` and the loop as it was. k are
-#  visited in the same order and everything below the GEMM is per-k, so the payload stream and the
-#  bracket order are unchanged.
 # ---- per-(k, q-tile) `iq` index build ------------------------------------------------------------
 #
 # Fill `iqs[1:nq]` with the index into `qpts` of `x_{k+q} - x_k` for outer k `ik` and every k+q of
@@ -626,14 +617,6 @@ function _loop_eph_over_k_and_kq_gpu(
     ndata_ekpR = nw * nbandk_max * nmodes
     nr_ep = length(model.epmat.irvec_next)
 
-    # ----- kR->kq GEMM strip width -----
-    # `nk_b` consecutive outer-k are stacked into one tall kR->kq GEMM, so the k-invariant phase tile
-    # is read nk_b times less and the GEMM's M dimension fills a cuBLAS tile. Shape-derived; it is 1
-    # (no strip, nothing changes) at large ndata or when the fused rotation kernel is not used. The
-    # M-target rule bounds the extra per-q staging (a wider `kRkq_ws.g`) at
-    # `16·(_KRKQ_GEMM_M_TARGET - ndata)` bytes for any model, so the q-tile below cannot collapse.
-    nk_b = _krkq_strip_width(; nw, nbandk_max, nmodes, nk_batch_max)
-
     # ----- memory-adaptive q-batch size (§7) -----
     # Every per-q staging buffer scales with the q-batch width, so cap it at what free device memory
     # allows (30% headroom for the batched drivers' recycled temporaries). The whole-run + per-k-batch
@@ -643,13 +626,12 @@ function _loop_eph_over_k_and_kq_gpu(
     # (Int, or nkq when nothing) stays a hard cap.
     per_point, committed = _outer_k_staging_bytes(; nw, nbandk_max, nmodes, nr_ep, nk, nkq,
         nq_grid = qpts.n, nk_batch_max, calculators,
-        ndata_epmat = epmat_dev.ndata, nr_epmat = epmat_dev.nr, nk_b, FT)
+        ndata_epmat = epmat_dev.ndata, nr_epmat = epmat_dev.nr, FT)
     nq_batch_cap = nq_batch_user === nothing ? nkq : min(nq_batch_user, nkq)
     nq_batch_max = plan_batch(backend, per_point, committed, nq_batch_cap; what = "outer-k")
     if verbosity > 0 && mpi_isroot()
         @info "GPU outer-k device memory: committed = $(round(committed / 1e9, digits = 2)) GB, " *
-              "$(round(per_point / 1e3, digits = 1)) kB/q; q-batch size = $nq_batch_max, " *
-              "kR->kq strip nk_b = $nk_b"
+              "$(round(per_point / 1e3, digits = 1)) kB/q; q-batch size = $nq_batch_max"
     end
 
     # ----- persistent workspace (allocated once, reused across all (k, q)) -----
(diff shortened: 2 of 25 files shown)


src/calculator/AbstractCalculator.jl
```diff
@@ -168,9 +169,12 @@ struct OuterIterationBatch end   # one batch of consecutive outer iterations (GP
 # (scope, loop-mode) combination the drivers fire on it, even when it wants a no-op — a missing
 # method is a loud error, never a silent skip (a silently-skipped bracket is a hard-to-find bug). The
 # combinations follow what the calculator `supports`: OuterIteration/SingleMode (CPU loops),
-# OuterIteration/BatchedMode (GPU outer-k per-k bracket; GPU outer-q per-q accumulator), and
-# OuterIterationBatch/BatchedMode (GPU outer-k per-batch). OuterIterationBatch/SingleMode is never
-# fired. A calculator that does nothing at a scope defines an explicit no-op (`= nothing`).
+# OuterIteration/BatchedMode (GPU outer-q per-q accumulator), and OuterIterationBatch/BatchedMode
+# (GPU outer-k per-batch). OuterIterationBatch/SingleMode is never fired, and neither is
+# OuterIteration/BatchedMode in the GPU OUTER-K loop: there the q-tile loop runs outside the k loop,
```
jaemolihm, line 174: really? did you check hubbard calculator?

src/common/kpoints.jl
```diff
@@ -245,6 +245,96 @@ function kpoints_create_subgrid(k::Kpoints, nsubgrid)
     Kpoints(new_n, new_vectors, new_weights, new_ngrid)
 end
 
+# Two dense tables of `prod(ngrid)` `Int`s are built in this file, and each has its own size
```
jaemolihm: First explain what we want to do (xk -> ik map for xk on a grid of size ngrid) and say there are two ways, hash dict and dense table. Then explain tradeoff (dict memory scale with n so low memory when very sparse, dense has fixed memory and lower when almost full (n ~ prod(ngrid)), access is faster for dense). Then say depending on the memory cap, we build either table or dict.

src/common/kpoints.jl
```diff
@@ -245,6 +245,96 @@ function kpoints_create_subgrid(k::Kpoints, nsubgrid)
...
+#   - waiver, `prod <= 4n` (includes every full grid at any size): `8·prod <= 32n` bytes against
+#     ~34n, so the table is never the bigger of the two. This is the clause that lets the retained
+#     index exceed 256 MB, and it is safe precisely because the Dict would have cost more.
+#   - density band, `4n < prod <= 8n`: up to `64n` bytes against ~34n, so up to ~1.9x the Dict —
+#     bounded in absolute terms by the 256 MB cap.
+#   - node floor, `prod <= 2^20`: the ratio is unbounded (5000 points on a (101,101,101) grid is an
+#     8.2 MB table for a ~170 kB Dict, ~48x), which is acceptable only because the floor itself
+#     bounds the table at 8 MB. That absolute bound is why this clause may ignore density at all.
+# Outside the waiver, exceeding the cap selects the Dict.
+const _DEDUP_TABLE_MAX_BYTES = 8 * 1024^3    # transient in `combine_kpoint_grids`
```
jaemolihm: Used only in one function, so move inside that

src/common/kpoints.jl
```diff
@@ -245,6 +245,96 @@ function kpoints_create_subgrid(k::Kpoints, nsubgrid)
...
+const _DENSE_INDEX_MAX_BYTES = 256 * 1024^2  # retained in `GridKpoints._dense_hash_to_ik`
```
jaemolihm: Used only in one function, move inside that. hardcode in test and comment which file has that number. Also why 256 and 1024^2? Why not 2^X? Oh you mean 256 MB, then comment like that

src/common/kpoints.jl
```diff
@@ -245,6 +245,96 @@ function kpoints_create_subgrid(k::Kpoints, nsubgrid)
...
 
+# Two dense tables of `prod(ngrid)` `Int`s are built in this file, and each has its own size
+# budget. They are deliberately two constants, not one: the de-duplication table is a transient
+# freed when `combine_kpoint_grids` returns, while the inverse index is retained for the lifetime of
+# every `GridKpoints` (and `split_kpoints` makes N of them), so the retained one is budgeted much
+# more tightly. Exceeding either budget selects the Dict implementation of the same map: slower,
+# O(number of points) memory, and identical results.
+#
+# The index budget is a soft cap, not a ceiling. A `GridKpoints` keeps at most one of the two
+# indices, but how far the dense table can exceed the Dict it replaces depends on which clause of
```
jaemolihm: one of the two "indices" -> "index maps"

src/common/kpoints.jl
```diff
@@ -245,6 +245,96 @@ function kpoints_create_subgrid(k::Kpoints, nsubgrid)
...
+# `_use_dense_index` admitted it (table 8 B per grid node, Dict ~34 B per point):
+#   - waiver, `prod <= 4n` (includes every full grid at any size): `8·prod <= 32n` bytes against
```
jaemolihm: If n >= prod(ngrid)/4, using dense table use less memory than Dict because Dict memory scale with n and use ~4 (34/8) times more memory per point than table

src/common/kpoints.jl
```diff
@@ -245,6 +245,96 @@ function kpoints_create_subgrid(k::Kpoints, nsubgrid)
...
+#     ~34n, so the table is never the bigger of the two. This is the clause that lets the retained
+#     index exceed 256 MB, and it is safe precisely because the Dict would have cost more.
+#   - density band, `4n < prod <= 8n`: up to `64n` bytes against ~34n, so up to ~1.9x the Dict —
```
jaemolihm: hard to understand this, revise

src/common/kpoints.jl
```diff
@@ -245,6 +245,96 @@ function kpoints_create_subgrid(k::Kpoints, nsubgrid)
...
+#     bounded in absolute terms by the 256 MB cap.
+#   - node floor, `prod <= 2^20`: the ratio is unbounded (5000 points on a (101,101,101) grid is an
```
jaemolihm: also revise

src/common/kpoints.jl
```diff
@@ -245,6 +245,96 @@ function kpoints_create_subgrid(k::Kpoints, nsubgrid)
...
+const _DEDUP_TABLE_MAX_BYTES = 8 * 1024^3    # transient in `combine_kpoint_grids`
+const _DENSE_INDEX_MAX_BYTES = 256 * 1024^2  # retained in `GridKpoints._dense_hash_to_ik`
+
+"""
+    _combine_kq_dedup_dense(BkS, BqS, sgn, ngrid_kq, shift_kq)
+De-duplicate `op(k, q)` over all (k, q) pairs with a dense `prod(ngrid_kq)` table indexed by the
+integer grid coordinates. Returns the unique k+q points in order of first appearance.
+
+The table is `[c1+1, c2+1, c3+1]` on `(ng1, ng2, ng3)`; de-duplication only needs *a* bijection
+from grid coordinates to table slots, so the order is free here. It is deliberately **not** the
```
jaemolihm: why not? why can't we make k3 the fast index? Then sequential k access becomes memory friendly

src/common/kpoints.jl
```diff
@@ -245,6 +245,96 @@ function kpoints_create_subgrid(k::Kpoints, nsubgrid)
...
+#     bounded in absolute terms by the 256 MB cap.
+#   - node floor, `prod <= 2^20`: the ratio is unbounded (5000 points on a (101,101,101) grid is an
+#     8.2 MB table for a ~170 kB Dict, ~48x), which is acceptable only because the floor itself
+#     bounds the table at 8 MB. That absolute bound is why this clause may ignore density at all.
+# Outside the waiver, exceeding the cap selects the Dict.
+const _DEDUP_TABLE_MAX_BYTES = 8 * 1024^3    # transient in `combine_kpoint_grids`
+const _DENSE_INDEX_MAX_BYTES = 256 * 1024^2  # retained in `GridKpoints._dense_hash_to_ik`
+
+"""
+    _combine_kq_dedup_dense(BkS, BqS, sgn, ngrid_kq, shift_kq)
```
jaemolihm, line 269: Document what BkS and BqS is

src/common/kpoints.jl
```diff
@@ -245,6 +245,96 @@ function kpoints_create_subgrid(k::Kpoints, nsubgrid)
...
+De-duplicate `op(k, q)` over all (k, q) pairs with a dense `prod(ngrid_kq)` table indexed by the
```
jaemolihm: op -> k + sgn * q

src/common/kpoints.jl
```diff
@@ -371,6 +427,50 @@ struct GridKpoints{T} <: AbstractKpoints{T}
     ngrid::NTuple{3,Int}
     shift::Vec3{T}
     _xk_hash_to_ik::Dict{Int,Int}
+    _dense_hash_to_ik::Vector{Int}
```
jaemolihm: why not 3d Array?

src/common/kpoints.jl
```diff
@@ -245,6 +245,96 @@ function kpoints_create_subgrid(k::Kpoints, nsubgrid)
...
+# Two dense tables of `prod(ngrid)` `Int`s are built in this file, and each has its own size
+# budget. They are deliberately two constants, not one: the de-duplication table is a transient
+# freed when `combine_kpoint_grids` returns, while the inverse index is retained for the lifetime of
+# every `GridKpoints` (and `split_kpoints` makes N of them), so the retained one is budgeted much
+# more tightly. Exceeding either budget selects the Dict implementation of the same map: slower,
+# O(number of points) memory, and identical results.
+#
+# The index budget is a soft cap, not a ceiling. A `GridKpoints` keeps at most one of the two
+# indices, but how far the dense table can exceed the Dict it replaces depends on which clause of
+# `_use_dense_index` admitted it (table 8 B per grid node, Dict ~34 B per point):
```
jaemolihm: Remove the following details, just refer to _use_dense_index for the actual criteria

src/common/kpoints.jl
```diff
@@ -371,6 +427,50 @@ struct GridKpoints{T} <: AbstractKpoints{T}
...
+
+# Every construction site that fills all fields goes through this, so the index policy lives in one
+# place: the dense table where `_use_dense_index` allows one, the Dict otherwise, never both. The
+# table is either empty or exactly `prod(ngrid)` long — the length that makes `_ik_from_hash`'s
+# `@inbounds` safe for any hash `_hash_xk` produces. `make_dict` supplies the Dict and is called
+# only when the Dict is the index that will be kept, so a caller that would have to build one from
+# scratch does not pay for it on a dense grid; callers holding one already can pass `() -> dict`.
+function _make_grid_kpoints(n, vectors::Vector{Vec3{T}}, weights, ngrid, shift,
+                            make_dict = () -> Dict{Int,Int}()) where {T}
+    _dense_hash_to_ik = _build_dense_index(n, vectors, ngrid, shift)
```
jaemolihm: if _use_dense_index else seems clearer. Also why can't this be part of GridKpoints constructor? You can have override flag to enforce dict or enforce dense table

src/common/kpoints.jl
```diff
@@ -447,27 +547,54 @@ function _hash_xk(xk::Vec3, ngrid, shift)
 end
 _hash_xk(xk, kpts::GridKpoints) = _hash_xk(xk, kpts.ngrid, kpts.shift)
 
+# Index of the k point with the given hash, 0 if there is none. `ik` is 1-based, so 0 is
+# unambiguous even though 0 is itself a legal hash (Γ on an unshifted grid).
+# PRECONDITION: `0 <= hash < prod(kpts.ngrid)`. Every hash built by `_hash_xk` satisfies it (it mods
+# each coordinate by `ngrid`), as does the equivalent integer-coordinate arithmetic in the GPU
+# outer-k loop. A caller that packs a hash by hand against a different `ngrid` violates it, and the
+# `@inbounds` below then reads out of bounds instead of reporting a miss.
```
jaemolihm: scary...

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6. Rewritten to your outline: what the map is (`xk -> ik` for `xk` on a grid of size `ngrid`), the two implementations, the tradeoff, then the cap decides which is built. The clause-by-clause breakdown is deleted and `_use_dense_index` is now named as the authority on the criteria.

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6 — now reads "keeps at most one of the two index maps".

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6. The waiver / density-band / node-floor bullets are deleted outright, not relocated; the block now ends with `_use_dense_index` as the authority on which map a grid gets.

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6, using your framing: at `n >= prod(ngrid)/4` the dense table already uses less memory than the Dict, because Dict memory scales with `n` and costs ~4x (34/8) the bytes per point that the table costs per node.

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6 — deleted along with the other two bullets rather than reworded.

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6 — deleted.

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6 — now a local `dedup_table_max_bytes = 8 * 1024^3  # 8 GB` inside `combine_kpoint_grids`, its only user.

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6 — now a local inside `_use_dense_index`, written `256 * 1024^2  # 256 MB, retained for the lifetime of the GridKpoints` so the magnitude is visible. The test hardcodes the value with a comment naming `_use_dense_index` in `src/common/kpoints.jl`, since it is no longer an importable constant.

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6 — documented as the integer grid coordinates of the k and q points on the common `ngrid_kq` grid, i.e. the `b` of `xk = shift + b ./ ngrid_kq`.

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6 — the docstring now writes it as `k + sgn * q`.

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6. The table is now `(ng3, ng2, ng1)` indexed `[c3+1, c2+1, c1+1]`, so `c3` is the unit-stride index.

It also has a payoff beyond locality: with those dims the linear index is exactly `hash + 1`, so the dedup table and `_dense_hash_to_ik` now share one layout. The paragraph claiming the two were deliberately different is gone. Output verified unchanged against the untouched Dict path — contents *and* first-appearance order.

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6 — the field is now `Array{Int,3}` sized `(ng3, ng2, ng1)`.

It is byte-identical to the flat vector it replaces: with those dims the column-major linear index of `[c3+1, c2+1, c1+1]` is exactly `hash + 1` for `hash = (c1*ng2 + c2)*ng3 + c3`, verified exhaustively on (2,3,4), (5,5,5), (4,7,3) and (150,150,150). `_ik_from_hash` still indexes linearly, so the lookup is unchanged — the 3-D type just makes the layout self-documenting instead of relying on a comment.

src/common/kpoints.jl
jaemolihm: Done in 9fbc3a6 — `_make_grid_kpoints` is now an explicit `if n == 0 ... elseif _use_dense_index(...) ... else ... end`.

Added the override too: `index = :auto | :dense | :dict` on the `GridKpoints(::Kpoints, ngrid)` constructor, threaded through `_use_dense_index` and `unfold_kpoints`/`fold_kpoints` (including their trivial-symmetry early returns, so the kwarg is never silently ignored); a bad symbol throws. The equivalence tests now use `index=:dict` to reach the fallback instead of relying on a sparse grid shape, which widened Dict-path coverage from one `(128,128,128)` grid to full/shifted/subset/irr plus both `sort!` branches.

src/common/kpoints.jl
jaemolihm: Agreed — it made the function partial: an out-of-domain hash read out of bounds instead of reporting a miss. Removed in 9fbc3a6; it now raises `BoundsError`.

Measured rather than assumed, since it runs ~1.11e9 times. The GPU wall A/B was unusable (another job shared the A100, 28-46 s swings within each arm), so I isolated the exact inner-loop body against the real Cu nk=150 `qpts` (nk=4918, nkq=226380, dense (150,150,150) table), interleaved, 1.114e8 pairs per rep:

| | ns/lookup |
|---|---|
| checked | 9.31 |
| `@inbounds` | 9.25 |

0.06 ns/lookup = **0.06 s over the full 1.11e9 lookups, ~0.2%** of the ~28 s loop. The bounds check is ~1 cycle against a guaranteed DRAM miss into a 27 MB table, so it is free. No `@boundscheck` escape hatch was added.

src/calculator/AbstractCalculator.jl
jaemolihm: Checked it directly — the comment is accurate for `HubbardDielectricCalculator` (`/mnt/home/jlihm/downfolding/Sr2RuO4/gpu/dev_hubbard_opt.jl`):

- `:145` declares `supports(..., OuterQLoop) = true` and there is no `OuterKLoop` method, so it defaults false. It consumes `EPData` (`:147`) and `EPDataKBatched` (`:148`) — outer-q.
- Its only brackets are `calculator_begin!/end!(::HubbardDielectricCalculator, ::OuterIteration, ctx)` (`:223`, `:241`), branching at runtime on `ctx.mode isa BatchedMode`. No `OuterIterationBatch` method.
- `run_eph_over_q_and_k.jl:509-510` builds `LoopContext(backend, BatchedMode(), iq)` and fires exactly that bracket.

So Hubbard is the comment's second listed case, `OuterIteration/BatchedMode` (GPU outer-q per-q accumulator), which the comment says **is** fired. The exclusion clause is scoped to the GPU **outer-k** loop, which Hubbard cannot enter. No wording change made.

Leaving this open since it was a question rather than a requested change — reopen the point if you read the wording differently.

One adjacent inconsistency found but not touched: `boltzmann_calculator.jl:244-245` says "the batched loop (`BatchedMode`) runs the default no-op", which contradicts this file's "There is NO no-op default". Harmless today (that combination is never fired for it), but worth its own thread if you want it fixed.

src/wannier/batched_fourier_core.jl
```diff
@@ -6,6 +6,43 @@ _alloc_array(parent::AbstractWannierObject, ::Type{S}, dims...) where {S} = zero
...
+(`(nr × 3)` real, row `ip` = `R_p`) and the crystal-coordinate points in `xkmat` (`(3 × nk)` real).
+All three arrays live on the same backend; the whole thing is a single broadcast (no scalar
+indexing, no scratch), so it runs unchanged on CPU and GPU.
+
+Stateless on purpose: the phase depends only on `(R_p, x)`, so a caller whose `x` list is
+loop-invariant can build it once and apply it many times (this is what the GPU outer-k e-ph loop
+does with its k+q-convention phase tile).
+"""
+function fourier_phase!(phase, irvec_mat, xkmat)
+    @views phase .= cispi.(2 .* (irvec_mat[:, 1] .* transpose(xkmat[1, :]) .+
```
jaemolihm: Do we need transpose ?

src/wannier/batched_fourier_core.jl
```diff
@@ -6,6 +6,43 @@ _alloc_array(parent::AbstractWannierObject, ::Type{S}, dims...) where {S} = zero
...
+    phase
+end
+
+"""
+    _irvec_matrix(irvec, proto, ::Type{T}) -> (nr × 3) real matrix
+
+R-vectors as an `(nr × 3)` real matrix on the backend of `proto` (a `WannierObject`), the layout
+[`fourier_phase!`](@ref) expects. Built on the host and copied over once.
+"""
+function _irvec_matrix(irvec, proto, ::Type{T}) where {T}
```
jaemolihm: _irvec_matrix_to_device ?

jaemolihm: _irvec_matrix_to_device ?

src/wannier_to_bloch_batched.jl
```diff
@@ -153,7 +153,7 @@ end
 #  wins on the GPU. Rotation matrices are stacked along the batch dimension.
 
 """
-    get_eph_RR_to_kR_batched!(ep_ekpR_all, itp_epmat::BatchedWannierInterpolator, ks, uks)
+    get_eph_RR_to_kR_batched!(ep_ekpR_all, itp_epmat::BatchedWannierInterpolator, ks, uks; kq_convention_phase=nothing)
```
jaemolihm: kq_convention_phase -> additional_phase

No reason to detail how this is being used for k+q convention, just explain what is being computed (what is shape of this kwarg and what is multiplied) and mention where this is used in one additional sentence

src/wannier_to_bloch_batched.jl
```diff
@@ -211,46 +230,69 @@ function KRtoKQWorkspace(gpu_array, ndata::Int, nbandkq::Int, nbandk::Int, nmode
 end
 
 """
+    get_eph_kR_to_kq_batched!(ep_kq_all, ep_kR::AbstractMatrix, phase::AbstractMatrix, u_phs, ukqs; ws=nothing)
```
jaemolihm, line 228: Is both methods needed?

src/wannier_to_bloch_batched.jl
```diff
@@ -371,23 +413,53 @@ function add_eph_dipole_batched!(eps, coeffs, ukqs, uks, mmats)
     eps
 end
 
+# Above this `nw*nmodes` the two rotation GEMMs are large enough that cuBLAS beats the CUDA
+# extension's fused kernel; see `_fused_eph_rot_kernel!` for the full rationale. It lives here
+# rather than in the extension so the base package documents the crossover the generic
+# `eph_apply_rotations!` docstring refers to. The crossover is hardware-dependent: the value was
+# tuned on one GPU and should be retuned elsewhere.
+const _FUSED_ROT_MAX_NWNM = 24
```
jaemolihm, line 406: Is this small-size opt needed? Benchmark for nk=150 with and without this fusing

src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -499,6 +500,38 @@ end
...
+#  is reused `nk_batch` times. Consequence for calculators: a k's work no longer finishes at a
+#  single point in the loop, so this loop fires only the `OuterIterationBatch` bracket — there is
+#  no per-k `OuterIteration` bracket in the batched outer-k loop.
+# ---- per-(k, q-tile) `iq` index build ------------------------------------------------------------
+#
+# Fill `iqs[1:nq]` with the index into `qpts` of `x_{k+q} - x_k` for outer k `ik` and every k+q of
+# the tile `qstart .+ (0:nq-1)`; the caller uploads it for the device phonon gather. A standalone
+# function so the `ik`/`qstart`/`nq` it works on are typed arguments rather than `Core.Box`-wrapped
+# captures of the three enclosing loop bodies.
+function _fill_iqs!(iqs, qpts, xkqs_int, xks_int, ik, qstart, nq)
```
jaemolihm, line 518: doesn't belong here, also explain the motivation more concisely

src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -648,36 +682,63 @@ function _loop_eph_over_k_and_kq_gpu(
         copyto!(ωq_all_dev, ωq_all_host)
     end
 
-    qs_batch  = Vector{Vec3{FT}}(undef, nq_batch_max)
-    iqs_batch  = Vector{Int}(undef, nq_batch_max)
-    iqs_batch_dev    = similar(epmat_dev.op_r, Int, nq_batch_max)
+    # Grid coordinates as (3 × n) real device matrices, uploaded once. Both phase builds below read
+    # them directly, so nothing on the phase path is staged on the host or copied H2D inside the loop.
+    xk_dev  = similar(epmat_dev.op_r, FT, 3, nk)
```
jaemolihm: Use proto variable not epmat_dev.op_r ?

src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -648,36 +682,63 @@ function _loop_eph_over_k_and_kq_gpu(
...
+    xkq_dev = similar(epmat_dev.op_r, FT, 3, nkq)
+    let xk_host = Matrix{FT}(undef, 3, nk), xkq_host = Matrix{FT}(undef, 3, nkq)
```
jaemolihm: can't we have _to_device for kpoints ? (or _kpoint_vectors_to_device )

src/wannier/batched_fourier_core.jl
jaemolihm: No — removed in `20088c4`. `xkmat` is `(3, n)`, so `xkmat[d, :]` dropped to a vector and then had to be transposed back into a row. `xkmat[d:d, :]` is already a `1 × n` row view:

```julia
phase .= cispi.(2 .* (irvec_mat[:, 1] .* xkmat[1:1, :] .+ ... ))
```

Bitwise identical (`Sᵢ` full-array FNV unchanged at Cu nk=150; `fourier_phase!` unit test green on CPU and CUDA). Not benchmarked separately — since the phase hoist this builds ~20 times per run rather than per pair, so it is no longer on a hot path.

src/wannier/batched_fourier_core.jl
jaemolihm: Renamed to `_irvec_matrix_to_device` in `20088c4` — it builds on the host and copies to the backend of `proto`, so the name should say so. Three call sites updated (`batched_fourier_core.jl:86`, `run_eph_over_k_and_kq.jl:699`, `test_gpu.jl:163`).

jaemolihm: Same as the other thread on this line — renamed to `_irvec_matrix_to_device` in `20088c4`, three call sites updated.

src/wannier_to_bloch_batched.jl
jaemolihm: Done in `20088c4`. Renamed `kq_convention_phase` → `additional_phase` (signature, body, docstring, driver, test, and `README_GPU.md:155`), and the docstring is now just what you asked for: the `(nr_ep × nk)` shape, that the child object is stored as `conj(P[ip,k]) · g(k, R_p)`, and one sentence naming the GPU outer-k loop as the user. The `# k+q convention` derivation block is gone.

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: Moved in `20088c4`. It had run together with the `iq`-build section header as one contiguous comment block — the bracket sentence now sits beside the "Supported / Not supported" contract text where the rest of the contract lives, cut to one sentence, with a blank line separating the loop header from the `iq` section.

One thing to flag: `run_eph_over_k_and_kq`'s own `nk_outer_batch_max` kwarg comment (`:22-25`) already states the same contract, so the moved sentence is now partly redundant with it. I left both — say the word and I will trim the kwarg one instead.

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: Done in `20088c4` — one `proto = epmat_dev.op_r` local, used by all 14 `similar` sites and the `KRtoKQWorkspace` call.

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: Done in `20088c4`, using your second name. New `_kpoint_vectors_to_device(backend, kpts)` in `src/common/kpoints.jl:823`, built on the existing `to_device(backend, x)` rather than a parallel mechanism; it returns the `(3 × kpts.n)` real device matrix and replaces the driver's 12-line `let` block with two lines. Probed on both backends (`Matrix{Float64}` and `CuArray`, equal to a hand-built reference).

src/wannier_to_bloch_batched.jl
jaemolihm: Benchmarked — it earns its keep by a wide margin, so **kept**. Cu nk=150, one binary with a temporary `Ref` gate on the extension's `if`, arms alternated after a discarded warm-up, one shared setup (`nk=4918, nkq=226380, nq=2896887`):

| arm | reps (s) | median |
|---|---|---|
| fused kernel (shipped; Cu `nw*nmodes` = 21 ≤ 24) | 30.943 / 31.188 | **31.07** |
| two cuBLAS GEMMs (gate forced off) | 47.037 / 46.953 | **47.00** |

**15.9 s = 33.9% of the two-GEMM wall (1.51×)** — far over the 10% bar, unlike the k-strip GEMM which I removed in `8e8b3cc` for failing it on reachability. The temporary gate was deleted before committing (`git diff` of `ext/` across `20088c4` is empty), and both numbers are now in the `_FUSED_ROT_MAX_NWNM` comment so this does not have to be asked again.

Two notes on the mechanism:

- Deleting it would **not** have moved the `g2 = |ep|²/(2ω)` fold elsewhere — the generic `eph_apply_rotations!` and the cuBLAS branch both implement it already. The fused version just does it from registers instead of in an extra full-array pass, which is part of the 15.9 s.
- The two paths are **not** bit-identical (`Sᵢ` FNV 15206647823536419969 fused vs 17918662762485534812 two-GEMM), so the cuBLAS path cannot serve as a bitwise reference for the fused one — cross-path comparisons need a tolerance.

Leaving this thread open for you to accept the keep decision.

src/wannier_to_bloch_batched.jl
jaemolihm: Checked, and **kept both** — but the premise is worth correcting: the second method has **no `src` caller at all**. The driver (`run_eph_over_k_and_kq.jl:817`) uses the direct `(ep_kR, phase)` form; the `qs` form is used only by **5 call sites in `test_gpu.jl`** (`:73, :102, :171, :352, :356`).

Reason for keeping rather than deleting: removing it puts a `fourier_phase!` build at each of those five sites, and two of them are inside `check_eph_partial_view`, where the sliced q-list would need a second build — so the tests would end up duplicating phase-building logic five times to delete one method. The docstring now says plainly which form is the driver's and which is the tests'.

Not resolving this one since it is a judgement call rather than a change you asked for — happy to delete it if you would rather the tests carry the phase build.

jaemolihm: Can we move the second method to test_gpu.jl and rename it?

src/wannier_to_bloch_batched.jl
jaemolihm: Sounds good, leave a comment that for nw=7, nmodes=3, fused is X% faster than cublas

src/common/kpoints.jl
```diff
@@ -371,16 +472,76 @@ struct GridKpoints{T} <: AbstractKpoints{T}
...
+    end
+    table
+end
+
+# Every construction site that fills all fields goes through this, so the index policy lives in one
+# place: exactly one of the two index maps is built, never both. `make_dict` supplies the Dict and
+# is called only when the Dict is the index that will be kept, so a caller that would have to build
+# one from scratch does not pay for it on a dense grid; callers holding one already can pass
+# `() -> dict`. `index` is forwarded to `_use_dense_index`.
+function _make_grid_kpoints(n, vectors::Vector{Vec3{T}}, weights, ngrid, shift,
```
jaemolihm: passing function argument is weird

src/common/kpoints.jl
```diff
@@ -604,6 +819,22 @@ function split_kpoints(kpts::GridKpoints, N)
...
+"""
+    _kpoint_vectors_to_device(backend, kpts::AbstractKpoints) -> (3 × kpts.n) real matrix
+
+Crystal coordinates of `kpts` as a `(3 × kpts.n)` real matrix on `backend`'s device — the layout the
+batched Fourier phase builds ([`fourier_phase!`](@ref)) read. Assembled on the host from the `Vec3`
+list and moved over once with [`to_device`](@ref).
+"""
+function _kpoint_vectors_to_device(backend, kpts::AbstractKpoints{T}) where {T}
+    xk_host = Matrix{T}(undef, 3, kpts.n)
+    for (ik, xk) in enumerate(kpts.vectors)
```
jaemolihm: can we do reinterpret ?

src/wannier/batched_fourier_core.jl
```diff
@@ -6,6 +6,44 @@ _alloc_array(parent::AbstractWannierObject, ::Type{S}, dims...) where {S} = zero
...
+Wannier → Bloch phase matrix `phase[ip, j] = exp(2πi R_p · x_j)` for the R-vectors in `irvec_mat`
+(`(nr × 3)` real, row `ip` = `R_p`) and the crystal-coordinate points in `xkmat` (`(3 × nk)` real).
+All three arrays live on the same backend; the whole thing is a single broadcast (no scalar
+indexing, no scratch), so it runs unchanged on CPU and GPU.
+
+Stateless on purpose: the phase depends only on `(R_p, x)`, so a caller whose `x` list is
+loop-invariant can build it once and apply it many times (this is what the GPU outer-k e-ph loop
+does with its k+q-convention phase tile).
+"""
+function fourier_phase!(phase, irvec_mat, xkmat)
```
jaemolihm, line 24: is this allocation free? Call this function separately and test. If free, good But comment that phase is (nr, nk) array. If not, just define rdotk array as temporary because the outer product via broadcast is a bit confusing.

src/wannier/batched_fourier_core.jl
```diff
@@ -6,6 +6,44 @@ _alloc_array(parent::AbstractWannierObject, ::Type{S}, dims...) where {S} = zero
...
+    phase
+end
+
+"""
+    _irvec_matrix_to_device(irvec, proto, ::Type{T}) -> (nr × 3) real matrix
+
+R-vectors as an `(nr × 3)` real matrix on the backend of `proto` (a `WannierObject`), the layout
+[`fourier_phase!`](@ref) expects. Built on the host and copied over once.
+"""
+function _irvec_matrix_to_device(irvec, proto, ::Type{T}) where {T}
```
jaemolihm: Move to WannierObject ? Also _irvec_to_device_matrix ? ( Same for k point vectors : _kpoints_to_device_matrix)

src/wannier_to_bloch_batched.jl
```diff
@@ -186,7 +194,13 @@ function get_eph_RR_to_kR_batched!(ep_ekpR_all::AbstractArray{Complex{T},3},
     C = similar(g, Complex{T}, nband, nw * M, nk)
     batched_gemm!('T', 'N', uks, reshape(gp, nw, nw * M, nk), C)        # C(k)=transpose(uk(k))*gp(k)
     out = permutedims(reshape(C, nband, nw, M, nk), (2, 1, 3, 4))       # (nw, nband, M, k)
-    copyto!(ep_ekpR_all, reshape(out, nw * nband * nmodes, nr_ep, nk))
+    out3 = reshape(out, nw * nband * nmodes, nr_ep, nk)
+    if additional_phase === nothing
+        copyto!(ep_ekpR_all, out3)
+    else
+        @assert size(additional_phase) == (nr_ep, nk)
+        ep_ekpR_all .= out3 .* conj.(reshape(additional_phase, 1, nr_ep, nk))
```
jaemolihm: Move conj to callsite

src/wannier_to_bloch_batched.jl
jaemolihm: Done in `ae94da4` — moved to `test/test_gpu.jl` as `kR_to_kq_from_qs!`, with a comment saying why it lives there. `src` now exposes one `get_eph_kR_to_kq_batched!` method (verified: `methods(...)` returns 1) and its docstring describes one thing. Five call sites moved; the two that already passed `(ep_kR, phase)` directly still call the `src` function, so both entry shapes stay covered.

(Reply is late — it was blocked by the 422 `user_id can only have one pending review` while the review was unsubmitted.)

src/wannier_to_bloch_batched.jl
jaemolihm: Comment added in `ae94da4`, next to `_FUSED_ROT_MAX_NWNM`: at nw=7, nmodes=3 the fused kernel is **1.51× faster than cuBLAS** — 34% less wall on the whole outer-k loop, 31.07 s vs 47.00 s at nk=150 on an A100-80GB. Given as both forms since "X% faster" is ambiguous. It also notes the two paths are **not** bit-identical (the fused one folds `g2` from registers rather than in a second full-array pass), because the cuBLAS path is the obvious reference to reach for and it will not match bitwise.

src/common/kpoints.jl
jaemolihm: Agreed — gone in `7f6c204`. `make_dict` is now `dict::Union{Nothing,Dict{Int,Int}} = nothing`: pass one if you hold it, leave it `nothing` and it is built inside, but only on the branch that keeps it, so a dense grid still never pays for a Dict.

Worth noting what the thunk was hiding: of the three callers, **two already held their Dict** (`sk_hash_dict`, `hash_dict_irr`) and were wrapping it in `() -> dict` purely to fit the signature. Only one was genuinely lazy, and that build now lives inside the function where it belongs.

src/common/kpoints.jl
jaemolihm: Yes — done in `7f6c204`. `Vec3{T}` is three contiguous `T`, so the host side is now a view, not a copy:

```julia
_kpoints_to_device_matrix(backend, kpts::AbstractKpoints{T}) where {T} =
    to_device(backend, reshape(reinterpret(T, kpts.vectors), 3, kpts.n))
```

One consequence I documented rather than hid: on `CPUBackend` `to_device` is the identity, so the result **aliases `kpts.vectors`**. Every caller is read-only (both are phase builds), so it is a free win there — but it is a genuine change in ownership, so tell me if you would rather it always copy.

src/wannier/batched_fourier_core.jl
jaemolihm: **Allocation-free** — `@allocated fourier_phase!(phase, irvec_mat, xkmat)` returns **0** on CPU with `nr=833, nk=64`. (Under CUDA it reports 6336 host-side bytes, which is kernel-launch bookkeeping, not an array.) So no `rdotk` temporary, and in `7f6c204` I added the comment you asked for:

```julia
# `phase` is `(nr, nk)`: entry `[ir, ik]` is `exp(2πi R[ir] · x[ik])`, an outer product over the
# `(nr × 3)` R-vectors and the `(3 × nk)` crystal coordinates. Verified allocation-free (`@allocated`
# returns 0 on CPU), so it needs no `rdotk` temporary.
```

src/wannier/batched_fourier_core.jl
jaemolihm: Both done in `7f6c204`. `_irvec_to_device_matrix` — renamed to your ordering and **moved to `src/wannier/WannierObject.jl`**, since it is about a `WannierObject`'s `irvec` and had no reason to sit in the Fourier core. `_kpoints_to_device_matrix` — renamed to match, and it now uses `reinterpret` per your other comment.

src/wannier_to_bloch_batched.jl
jaemolihm: Done in `7f6c204`. `get_eph_RR_to_kR_batched!` now multiplies by the `additional_phase` it is handed; the two callers conjugate their own — the driver (`P_k[:, 1:nk_batch] .= conj.(...)` right after the build) and the k+q-convention test. The callee no longer encodes one caller's convention, and the docstring says the loop passes `conj(exp(2πi R_p · x_k))` rather than burying it.

This one initially **failed** the `k+q convention identity` test, which builds its own `P_k` — fixed by conjugating at that call site too. Good failure: it means that test was checking the convention rather than round-tripping it.

jaemolihm:
Note on commit hashes: I force-pushed to correct `7f6c204`, which had accidentally committed 12 working documents (`plans/` plus `CALCULATOR_ARCHITECTURE_REVIEW.md`, `KPOINTS_HASH_PLAN.md`, `RT_BTE_PLAN.md`, `WANNIER90_LOADER_PLAN.md`, 6872 lines) that were never meant to be tracked. The code content is unchanged — same six source/test files, same 21602 passing tests.

**`7f6c204` is now `d695a44`.** Replies above that cite `7f6c204` mean `d695a44`; `ae94da4` and everything earlier are unaffected. All five paths are now in `.git/info/exclude` so it cannot recur.

src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -602,31 +636,31 @@ function _loop_eph_over_k_and_kq_gpu(
               "$(round(per_point / 1e3, digits = 1)) kB/q; q-batch size = $nq_batch_max"
     end
 
-    itp_ep_ekpR = BatchedWannierInterpolator(ep_ekpR_dev; batch_size = nq_batch_max)
-
     # ----- persistent workspace (allocated once, reused across all (k, q)) -----
     # All device staging is sized to the full batch and used as plain CuArrays (not
     # batch-sliced views), so the batched drivers' reshape/cuBLAS calls stay on dense arrays.
+    # `proto` is the `similar` template every device buffer below is allocated from.
+    proto = epmat_dev.op_r
```
jaemolihm: Why not backend ?

src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -701,101 +756,88 @@ function _loop_eph_over_k_and_kq_gpu(
...
+                # device SubArray view instead would fall back to scalar indexing); the device gather
+                # then reads only `view(iqs_batch_dev, rng_q)`, so the untouched tail is never used.
+                # `@inbounds`: bounds-checking a device INDEX ARRAY costs a Bool map+reduce kernel and
+                # a D2H round trip of the result per (k, tile); `iq` is already validated on the host
+                # (the guard in `_fill_iqs!`), so the device-side re-check is pure overhead.
+                _fill_iqs!(iqs_batch, qpts, xkqs_int, xks_int, ik, qstart, nq_batch)
+                copyto!(iqs_batch_dev, 1, iqs_batch, 1, nq_batch)
+                iqs_used = view(iqs_batch_dev, rng_q)
+                @inbounds @views uphs_dev[:, :, rng_q] .= uph_all_dev[:, :, iqs_used]
+                @inbounds @views ωq_dev[:, rng_q]      .= ωq_all_dev[:, iqs_used]
```
jaemolihm, line 813: Can these steps wrapped in a function, instead of only _fill_iqs! wrapped? Tell me if this is not good idea

src/common/kpoints.jl
```diff
@@ -371,16 +472,76 @@ struct GridKpoints{T} <: AbstractKpoints{T}
...
+    end
+    table
+end
+
+# Every construction site that fills all fields goes through this, so the index policy lives in one
+# place: exactly one of the two index maps is built, never both. Pass `dict` if you already hold the
+# `hash => ik` map; leave it `nothing` and one is built here, but only on the branch that keeps it,
+# so a dense grid never pays for a Dict. `index` is forwarded to `_use_dense_index`.
+function _make_grid_kpoints(n, vectors::Vector{Vec3{T}}, weights, ngrid, shift,
+                            dict::Union{Nothing,Dict{Int,Int}} = nothing; index = :auto) where {T}
```
jaemolihm: is passing dict just to save cost of dict regeneration? This will be small

src/common/kpoints.jl
```diff
@@ -371,16 +472,76 @@ struct GridKpoints{T} <: AbstractKpoints{T}
...
+        table[_hash_xk(vectors[ik], ngrid, shift) + 1] = ik
+    end
+    table
+end
+
+# Every construction site that fills all fields goes through this, so the index policy lives in one
+# place: exactly one of the two index maps is built, never both. Pass `dict` if you already hold the
+# `hash => ik` map; leave it `nothing` and one is built here, but only on the branch that keeps it,
+# so a dense grid never pays for a Dict. `index` is forwarded to `_use_dense_index`.
+function _make_grid_kpoints(n, vectors::Vector{Vec3{T}}, weights, ngrid, shift,
```
jaemolihm: Can't this be GridKpoints constructor?

src/common/kpoints.jl
```diff
@@ -439,50 +599,99 @@ end
...
+        end
+    else
+        for (ik, xk) in enumerate(k.vectors)
+            k._dense_hash_to_ik[_hash_xk(xk, k) + 1] = ik
+        end
     end
     k
 end
 
+# The two indices are derived from (vectors, ngrid, shift), so they are not compared.
```
jaemolihm: Remove

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: You're right, and it was worse than a style point — `alloc(backend, T, dims...)` **already exists** in `src/common/gpu_utils.jl`, and `run_eph_over_q_and_k` was already using it. My `proto` local was a second, redundant mechanism for the same thing. Fixed in `478f27e`: all 12 device buffers now come from `alloc(backend, ...)`, so the backend is the single authority on where "device" is, and the two drivers agree.

(`KRtoKQWorkspace` still takes an array template rather than a backend — that is its own signature, not this loop's business.)

src/common/kpoints.jl
jaemolihm: Measured, and you're right: `Dict(_hash_xk.(...) .=> 1:n)` is **0.6 ms at n = 11 525 and 29.5 ms at n = 535 998** — and only on the Dict branch, which realistic grids do not take. Not worth threading a pre-built Dict through every caller.

Dropped in `478f27e`. Two of the three callers were passing a Dict they happened to hold; the rest passed nothing. It is now always built inside, on the branch that keeps it.

src/common/kpoints.jl
jaemolihm: Yes — done in `478f27e`. It is now `GridKpoints(n, vectors, weights, ngrid, shift; index)` with a `GridKpoints{T}` method behind it, and `_make_grid_kpoints` is gone.

Dropping the `dict` parameter (your other comment) is what made this clean: a 5-argument constructor rather than a 6-argument one with a nothing-able trailing parameter. There is also a non-parametric forwarder, because two call sites (`unfold_kpoints`, `fold_kpoints`) have no bound `T`.

src/common/kpoints.jl
jaemolihm: Removed in `478f27e`. Worth saying why it deserved removing beyond being unnecessary: it compared `vectors` and `weights` with `≈`, and `n` — an `Int` — with `≈` too, which is the wrong semantics for `==`.

Two **live** call sites compared `GridKpoints` and now compare the defining fields explicitly: `test/boltzmann/test_hdf5.jl:108` (HDF5 round-trip) and `test/test_kpoints.jl:347` (deprecated alias vs `combine_kpoint_grids`). I found the second only after the first test run failed — my initial grep pattern missed it. An exhaustive sweep over all 17 live test files confirms there are no others.

`test_unfold.jl` has three more, updated for consistency but **unverifiable here**: it is commented out at `runtests.jl:18` and its `data_cubicBN` fixtures are absent from the checkout.

Tests: 21702 pass.

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: Not a good idea as a plain function, but there is a good version of it — here is the reasoning so you can judge.

The block needs **~26 live values**: `iqs_batch`, `iqs_batch_dev`, `uphs_dev`, `uph_all_dev`, `ωq_dev`, `ωq_all_dev`, `epkq_dev`, `g2_dev`, `ep_ekpR_all`, `P_kq`, `ukqs_used`, `kRkq_ws`, `ibandk_offsets`, `xkqs_int`, `xks_int`, `qpts`, `ik`, `ik_ind`, `qstart`, `nq_batch`, `rng_q`, `ikqs_used`, `backend`, `iks_batch`, `nk_batch_max`, `calculators`. A 26-argument function reads worse than the inline block, and a closure would reintroduce the `Core.Box` the `_setup`/`_loop` split exists to avoid. `_fill_iqs!` is wrappable precisely because it needs only 7.

The version that works is to bundle the device buffers into one struct and pass that, which cuts it to ~6 arguments and makes the wrap natural. That is already anticipated by the `TODO` on `KRtoKQWorkspace` ("merge this with the other Wannier→Bloch scratch ... into a single shared workspace so a whole computation allocates its device scratch once"). I would rather do that as its own change than inside this PR — say the word if you want it here.

Leaving this open for you to decide.

src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -648,37 +686,54 @@ function _loop_eph_over_k_and_kq_gpu(
...
+    # The two Fourier phase matrices of the k+q convention (see `get_eph_RR_to_kR_batched!`):
+    #   P_k [ip, k] = exp(2πi R_p · x_k)      — conjugated into g(k, R_ep) once per outer-k batch
+    #   P_kq[ip, j] = exp(2πi R_p · x_{k+q_j}) — the kR->kq phase, INDEPENDENT of the outer k
+    # so one built P_kq tile serves every k of the batch. That reuse factor is `nk_batch`, i.e.
+    # `nk_outer_batch_max`: lowering that cap shrinks this saving proportionally.
+    # Not a `TiledDeviceOutput`: that tiles the OUTPUT side over the outer-state axis and owns a
+    # host mirror plus a per-batch D2H, whereas P_kq is an input-side, q-indexed,
+    # write-once-read-many device buffer with no host side and no D2H at all.
+    irvecp_mat = _irvec_to_device_matrix(model.epmat.irvec_next, epmat_dev, FT)
+    P_k  = alloc(backend, Complex{FT}, nr_ep, nk_batch_max)
```
jaemolihm: Call it P_mk ?

src/wannier/WannierObject.jl
```diff
@@ -132,3 +132,21 @@ function get_next_wannier_object(parent :: WannierObject{T}; ndata_child = nothi
...
 end
+
+
+"""
+    _irvec_to_device_matrix(irvec, proto, ::Type{T}) -> (nr × 3) real matrix
+
+R-vectors of a `WannierObject` as an `(nr × 3)` real matrix on the backend of `proto`, the layout
+[`fourier_phase!`](@ref) expects. Built on the host and copied over once.
+"""
+function _irvec_to_device_matrix(irvec, proto, ::Type{T}) where {T}
```
jaemolihm: proto -> device

src/wannier/WannierObject.jl
```diff
@@ -132,3 +132,21 @@ function get_next_wannier_object(parent :: WannierObject{T}; ndata_child = nothi
...
+    nr = length(irvec)
+    irvec_host = Matrix{T}(undef, nr, 3)
+    for ir in 1:nr, d in 1:3
+        irvec_host[ir, d] = irvec[ir][d]
+    end
+    irvec_mat = _alloc_array(proto, T, nr, 3)
+    copyto!(irvec_mat, irvec_host)
```
jaemolihm: can we do to_device(...) ?

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: Done in `86f984e` — `P_mk`, following the index-layout convention (`m` = R-vector index, then `k`), as with `ep_mjν`.

I renamed `P_kq` → `P_mkq` at the same time: `P_mk` sitting next to `P_kq` would read worse than either convention applied consistently. 22 occurrences across the driver, `test_gpu.jl`, `README_GPU.md` and the byte-accounting comments in `eph_device_staging.jl`.

src/wannier/WannierObject.jl
jaemolihm: Renamed in `86f984e`, and the docstring now says what it is rather than leaving `proto` to be inferred: the `WannierObject` whose storage the result should match.

src/wannier/WannierObject.jl
jaemolihm: Checked, and no — with the reason now written into the docstring rather than left implicit:

1. `to_device(backend, x)` takes a **backend**, and this is reached from `BatchedFourierCore(parent)`, which has only the object in scope — no backend.
2. More substantively, `_alloc_array` deliberately keeps a **`DiskWannierObject`**'s array on the **host** (it has no in-memory `op_r`, see its definition in `batched_fourier_core.jl:5-6`). A backend cannot express "host for disk-backed objects, device for in-memory ones", so switching would silently break that fallback.

I did tidy the body while there — the four-line host fill is a comprehension, and `_alloc_array` + `copyto!` collapse to one line since `copyto!` returns its destination:

```julia
function _irvec_to_device_matrix(irvec, device, ::Type{T}) where {T}
    nr = length(irvec)
    irvec_host = T[irvec[ir][d] for ir in 1:nr, d in 1:3]
    copyto!(_alloc_array(device, T, nr, 3), irvec_host)
end
```

Verified on both backends against a hand-built reference: host gives `Matrix{Float64}`, device gives `CuArray`, both equal.

src/calculator/run_eph_over_k_and_kq.jl
jaemolihm: You're right — I misread the name. I took the `m` as the R-vector index of the `ep_mjν` layout convention and renamed `P_kq` → `P_mkq` for symmetry. `P_mk` means the phase at **minus** `x_k`, and `P_kq` is not conjugated, so there is no minus in it. **`P_mkq` reverted to `P_kq` in `3261231`.**

Understanding the name also improved the code. `P_mk` was being built as `exp(2πi R_p·x_k)` and then conjugated in place; it can just be built at `−x_k`:

```julia
mxk_dev = -_kpoints_to_device_matrix(backend, kpts)   # once, at setup
...
@views fourier_phase!(P_mk[:, 1:nk_batch], irvecp_mat, mxk_dev[:, iks_batch])
```

That drops one `conj` pass over `(nr_ep × nk_batch)` per outer-k batch, and `xk_dev` had no other consumer. Bitwise identical, not just equal: FP negation is exact and the dot product is sign-symmetric, so the `cispi` argument is exactly negated — verified `conj(fourier_phase!(x)) == fourier_phase!(-x)` on CPU **and** CUDA at nr=833, nk=64.

Side benefit on the aliasing question from the other thread: `-view` allocates, so this call site no longer aliases `kpts.vectors`.

Tests: 21606 pass.

jaemolihm: Agreed, `-_kpoints...` juxtaposes the minus with the leading underscore and reads badly. Changed to your form in `f4be28d`:

```julia
mxk_dev = -1 .* _kpoints_to_device_matrix(backend, kpts)
```

Bitwise identical to `-x`, signed zeros included — checked `reinterpret(UInt64, -1 .* x) == reinterpret(UInt64, -x)` on CPU and CUDA over `[0.0, -0.0, 0.25, -0.375, 1e-300, 1.0]`.

I also added a note on why it must stay **out-of-place**: the tidier-looking `mxk_dev .*= -1` would be a bug, because on `CPUBackend` `_kpoints_to_device_matrix` returns a `reinterpret` view onto `kpts.vectors` — negating in place would corrupt the k-points themselves.

PR #12, opened 2026-08-03
perf(kpoints): 26x faster k+q grid de-duplication (mark-then-collect, threaded, pre-reduced)

jaemolihm:
Speeds up `combine_kpoint_grids` by **28.5×** at Cu nk=200: **76.86 → 2.69 s**.

Re-baselined on `workergpu053` (A100-SXM4-80GB, Xeon Platinum 8358, 16 threads) with the pre-change implementation and the new one alternating **in one session off a single model load**, so the ratio is not taken across sittings or hosts. Real grids: `kpts.n = 11525`, `kqpts.n = 535998`, 6.18e9 pairs, `qpts.n = 6870740`; the two results compare equal on vectors, weights, `ngrid` and `shift`.

This is CPU-only host work in *setup*, so it benefits non-GPU runs equally. It was found while profiling a GPU BTE run: at nk=200 it was the single largest setup cost, larger than either electron-state filter.

### Base branch

Stacked on `gpubte-opt` (#11), not `main` — the pre-reduction helpers `_grid_coords_reduced` / `_wrap_reduced` are introduced there by the outer-k `iq` build, and this commit reuses them in 12 places. The diff shown here is only the k-point work. Merge #11 first.

### Where the time went

`combine_kpoint_grids` ran nk × nkq = **6.18e9 pairs**, each doing three `mod`s and a random probe into a `prod(ngrid)` table. Measured split before the change (nk=150): pair loop 11.65 s, everything else 0.66 s.

Three levers, all on the pair loop. Per-lever times below are at nk=150 on `workergpu069` (also A100-80GB) — a different sitting from the nk=200 headline, so read the ratios, not the absolutes:

| step | time | ns/pair |
|---|---|---|
| baseline (`Int` table, `mod`) | 11.66 s | 10.47 |
| + pre-reduce coords, fold with `_wrap_reduced` | 4.30 s | 3.87 |
| + mark-then-collect into an `Array{Bool}` | 2.97 s | 2.67 |
| + `@threads` over `BqS` (16 threads) | 0.25 s | 0.22 |

Pre-reducing was the biggest single lever (2.7×) — bigger than the cache-footprint cut (1.45×), which is the reverse of what was expected going in.

- **Pre-reduce** `BkS`/`BqS` into `[0, ngrid)` once per point, so `bk + sgn*bq` lands in `(-ng, 2ng)` and needs only a compare-and-add rather than a runtime 64-bit integer division.
- **Mark-then-collect.** Pass 1 only marks occupied slots, in an `Array{Bool}` (one byte per element) rather than an `Array{Int}` of indices — the probe is random-access, so 8× less table (8 MB vs 64 MB at 200³, against a 48 MB L3) is 8× more of it cache-resident. Pass 2 sweeps in index order and emits the points, which also sizes `xkqs` up front instead of growing it.
- **Thread pass 1.** Legal because marking is idempotent (every writer stores the same `true`, so no synchronization) and because collect order is discarded downstream: `combine_kpoint_grids` sorts on integer grid coordinates — a total order over distinct points — and renumbers the index.

### Correctness

The returned `GridKpoints` is **exactly equal** — vectors, weights, `ngrid`, `shift`, and the populated index — over 275 randomized cases spanning both ops, non-cubic grids, differing input densities, random shifts and random grid subsets. Pass 2 rebuilds each point with the same `Vec3(c1/ng1, c2/ng2, c3/ng3) + shift_kq` expression on the same integers, so no rounding moves.

Verified independently of the author: `test_kpoints.jl` + `test_iq_build.jl` = **21422 tests passing**, and the 28.5× re-measured from scratch against the pre-change implementation on the real grids, as described at the top.

### Review round — `a4a5a07`

- **`fk`/`fq` gone.** `ngrid_kq` is a multiple of both input grids, so `_grid_coords_reduced(xk, ngrid_kq, shift)` reaches the common-grid coordinates in one step. Consequence: it rounds at `ngrid_kq` rather than at the coarser input grid, needing each point on-grid to `0.5/ngrid_kq` instead of `0.5/ngrid` — 13 orders of margin at 1e-16 relative error against `ngrid_kq = 200`. This left `_grid_coords_reduced(::Vec3{Int}, ngrid)` with no caller, so it is deleted.
- **Mark table is `Array{Bool}`**, not `Array{UInt8}` of `0x01`. `Bool` is one byte, so the footprint and the race-free byte stores are identical; `BitArray` is what would race, and the docstring now says so.
- **`@inbounds` audited, and measured rather than assumed.** Kept in `_mark_kq_pairs!` — worth **1.43×** on that pass (1.048 vs 1.498 s), ~17% of the function — with the justification written next to it. Dropped from the pass-2 sweep, where it cost nothing measurable.
- **`_wrap_reduced` now checks its range in a `@boundscheck`.** Outside `(-ng, 2ng)` it silently returned an out-of-range value, which under `@inbounds` is a segfault rather than a wrong answer. The `Dict` path and any future caller get the check; the hot loops elide it on the strength of `_check_reduced_coords`.

### Sharp edge worth reviewing

The precondition behind the kept `@inbounds` is what makes it safe, so it is worth checking that layering directly: `_check_reduced_coords` runs once per list at entry (O(nk+nkq), unmeasurable) and establishes `0 <= b < ng` for every entry; `_wrap_reduced` then cannot leave `0:ng-1`. It is not theoretical — the pre-existing `test_kpoints.jl` dedup test passed raw `rand(-6:6,3)` coordinates and killed the Julia process before the guard existed. Reviewers should satisfy themselves that no caller can reach the folds without passing that guard.

### Known behaviour change

The `Dict` fallback gets the pre-reduction only — its insert cannot be threaded and it has no table to sweep — so it now returns **first-appearance order** where the dense path returns **grid order**. The two still agree as sets, which is what the caller's `sort!` needs, and the test asserts set equality rather than order for that path.

### Follow-up, not in this PR

The trailing `sort!` now dominates: at nk=200 the split is coords / pair loop / `GridKpoints`+`shift_center!`+`sort!` ≈ 0.09 / 1.14 / 1.76 s. That is where any further work on this function goes.

### DIFF
--- a/src/common/kpoints.jl
+++ b/src/common/kpoints.jl
@@ -278,12 +278,13 @@ that a sweep over consecutive `c3` walks the table sequentially. With those dims
 of a slot is exactly `hash + 1` for the `(c1*ng2 + c2)*ng3 + c3` packing, so this table and
 `GridKpoints._dense_hash_to_ik` share one layout.
 
-Two passes rather than one. Pass 1 only MARKS which slots occur, in a `Vector{UInt8}` instead of
-the `Array{Int,3}` of indices the single-pass version needed: the pair loop's access into the table
-is a random probe, so at 8 B per node it missed cache on nearly every pair, and 1 B per node is 8×
-more of the grid resident (8 MB vs 64 MB at 200³). Marking is also idempotent — every writer stores
-the same 1 — which is what lets the pass be threaded with no synchronization. (A `BitArray` would
-be smaller still but its writes race at word granularity.)
+Two passes rather than one. Pass 1 only MARKS which slots occur, in an `Array{Bool}` instead of the
+`Array{Int}` of indices the single-pass version needed: the pair loop's access into the table is a
+random probe, so at 8 B per node it missed cache on nearly every pair, and `Bool` is 1 B per node,
+putting 8× more of the grid in cache (8 MB vs 64 MB at 200³). Marking is also idempotent — every
+writer stores the same `true` — which is what lets the pass be threaded with no synchronization.
+(A `BitArray` would be smaller still but its writes race at word granularity; `Array{Bool}` stores
+one byte per element, so its writes do not.)
 Pass 2 sweeps the table in index order and emits the points. That discards first-appearance order,
 which is safe because `combine_kpoint_grids` sorts the result on the integer grid coordinates — a
 total order, the points being distinct — and `sort!(::GridKpoints)` renumbers the index afterwards.
@@ -291,13 +292,13 @@ total order, the points being distinct — and `sort!(::GridKpoints)` renumbers 
 function _combine_kq_dedup_dense(BkS, BqS, sgn, ngrid_kq, shift_kq::Vec3{T}) where {T}
     ng1, ng2, ng3 = ngrid_kq
     _check_reduced_coords(BkS, BqS, ngrid_kq)
-    seen = zeros(UInt8, ng3, ng2, ng1)
+    seen = zeros(Bool, ng3, ng2, ng1)
     _mark_kq_pairs!(seen, BkS, BqS, sgn, ng1, ng2, ng3)
 
-    xkqs = Vector{Vec3{T}}(undef, count(!iszero, seen))
+    xkqs = Vector{Vec3{T}}(undef, count(seen))
     ikq = 0
-    @inbounds for c1 in 0:(ng1 - 1), c2 in 0:(ng2 - 1), c3 in 0:(ng3 - 1)
-        iszero(seen[c3 + 1, c2 + 1, c1 + 1]) && continue
+    for c1 in 0:(ng1 - 1), c2 in 0:(ng2 - 1), c3 in 0:(ng3 - 1)
+        seen[c3 + 1, c2 + 1, c1 + 1] || continue
         ikq += 1
         xkqs[ikq] = Vec3(c1 / ng1, c2 / ng2, c3 / ng3) + shift_kq
     end
@@ -320,9 +321,14 @@ end
 # typed arguments rather than `Core.Box`. Partitioned over `BqS` (the longer list in the k+q use),
 # each chunk sweeping the whole `BkS` — chunks share `seen` but only ever write the same value into
 # a slot, so no synchronization is needed and the result does not depend on the partition.
-function _mark_kq_pairs!(seen::Array{UInt8,3}, BkS::Vector{Vec3{Int}}, BqS::Vector{Vec3{Int}},
+function _mark_kq_pairs!(seen::Array{Bool,3}, BkS::Vector{Vec3{Int}}, BqS::Vector{Vec3{Int}},
                          sgn::Int, ng1::Int, ng2::Int, ng3::Int)
     @threads for iq in eachindex(BqS)
+        # `@inbounds` here is load-bearing for performance, not tidiness: this block runs
+        # nk*nkq ~ 1e9 times. It is safe because `_check_reduced_coords` has already established
+        # `0 <= b < ng` for every entry of both lists, so `_wrap_reduced` returns a value in
+        # `0:ng-1` on all three axes and every index below is in range. It also elides
+        # `_wrap_reduced`'s own `@boundscheck`, whose precondition that guard is what establishes.
         @inbounds begin
             bq = BqS[iq]
             q1, q2, q3 = sgn * bq[1], sgn * bq[2], sgn * bq[3]
@@ -330,7 +336,7 @@ function _mark_kq_pairs!(seen::Array{UInt8,3}, BkS::Vector{Vec3{Int}}, BqS::Vect
                 c1 = _wrap_reduced(bk[1] + q1, ng1)
                 c2 = _wrap_reduced(bk[2] + q2, ng2)
                 c3 = _wrap_reduced(bk[3] + q3, ng3)
-                seen[c3 + 1, c2 + 1, c1 + 1] = 0x01
+                seen[c3 + 1, c2 + 1, c1 + 1] = true
             end
         end
     end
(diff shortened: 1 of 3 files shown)


src/common/kpoints.jl
```diff
@@ -373,10 +416,14 @@ function combine_kpoint_grids(kpts, qpts, op, ngrid_kq)
...
+    # Both lists are reduced into `0:ng-1` here, once per point, so that the O(nk·nkq) fold is
+    # `_wrap_reduced` (a compare-and-add) rather than three runtime 64-bit integer divisions.
     fk = ngrid_kq .÷ kpts.ngrid
     fq = ngrid_kq .÷ qpts.ngrid
-    BkS = [Vec3(round.(Int, (xk - shift_k).data .* kpts.ngrid) .* fk) for xk in kpts.vectors]
-    BqS = [Vec3(round.(Int, (xq - shift_q).data .* qpts.ngrid) .* fq) for xq in qpts.vectors]
+    BkS = [_grid_coords_reduced(Vec3(round.(Int, (xk - shift_k).data .* kpts.ngrid) .* fk), ngrid_kq)
+           for xk in kpts.vectors]
+    BqS = [_grid_coords_reduced(Vec3(round.(Int, (xq - shift_q).data .* qpts.ngrid) .* fq), ngrid_kq)
+           for xq in qpts.vectors]
```
jaemolihm: The logic here is confusing. Can we avoid defining fk, fq ?

src/common/kpoints.jl
```diff
@@ -267,54 +268,96 @@ end
...
+                         sgn::Int, ng1::Int, ng2::Int, ng3::Int)
+    @threads for iq in eachindex(BqS)
+        @inbounds begin
+            bq = BqS[iq]
+            q1, q2, q3 = sgn * bq[1], sgn * bq[2], sgn * bq[3]
+            for bk in BkS
+                c1 = _wrap_reduced(bk[1] + q1, ng1)
+                c2 = _wrap_reduced(bk[2] + q2, ng2)
+                c3 = _wrap_reduced(bk[3] + q3, ng3)
+                seen[c3 + 1, c2 + 1, c1 + 1] = 0x01
```
jaemolihm: what is 0x01 ? Can't we use bool ?

src/common/kpoints.jl
```diff
@@ -1,6 +1,7 @@
 # Data and functions for k points
```
jaemolihm, line 1: Please make sure all @inbounds are optimization. One must know there is  no oob access. Ideally, just remove all @inbounds except performance critical ones, and leave justification when @inbounds is there.

src/common/kpoints.jl
jaemolihm: Done in `a4a5a07`. `ngrid_kq` is a multiple of both input grids, so every point sits on it exactly and one call gets there directly:

```julia
BkS = [_grid_coords_reduced(xk, ngrid_kq, shift_k) for xk in kpts.vectors]
BqS = [_grid_coords_reduced(xq, ngrid_kq, shift_q) for xq in qpts.vectors]
```

`fk`/`fq` are gone and the surrounding comment is cut to five lines. One honest consequence: the one-step form rounds at `ngrid_kq` instead of at the (coarser) input grid, so it needs each point on-grid to `0.5/ngrid_kq` rather than `0.5/ngrid` — at ~1e-16 relative error against `ngrid_kq = 200` that is 13 orders of margin, so it is not a practical constraint.

This also left `_grid_coords_reduced(::Vec3{Int}, ngrid)` with no caller, so I deleted it; the two test sites now say `mod.(b, ngrid)` directly. That also answers the "could this be one function with `shift` defaulting to zero" question — the two methods were doing different things (one converts fractional coordinates, one only reduces integers), so merging them would have made eltype dispatch load-bearing for a semantic difference. Deleting the redundant one is the better fix.

src/common/kpoints.jl
jaemolihm: Yes — you are right, and `Bool` is now what it uses (`a4a5a07`): `zeros(Bool, …)`, `seen[…] = true`, `count(seen)`, `seen[…] || continue`.

I had reached for `UInt8` out of a conflation on my part: what races at word granularity is `BitArray`, which is bit-packed. `Array{Bool}` is one byte per element, so it has exactly the same 8 MB footprint at 200³ and the same race-free byte stores as `UInt8` — no reason to spell it `0x01`. Re-measured after the change: `combine_kpoint_grids` 2.69 s at Cu nk=200, unchanged. The docstring now states the `Bool` vs `BitArray` distinction so the next reader does not have to rediscover it.

src/common/kpoints.jl
jaemolihm: Audited all three, measured each rather than guessing (`a4a5a07`).

**Kept — `_mark_kq_pairs!`.** Measured worth **1.43×** on that pass: 1.048 s with vs 1.498 s without, Cu nk=200, 6.18e9 pairs, 16 threads on workergpu053. That is ~0.45 s of a 2.69 s function, ~17%, so it is load-bearing rather than tidiness. It now carries the justification you asked for: safety rests on `_check_reduced_coords` having already established `0 <= b < ng` for every entry of both lists, which makes `_wrap_reduced` return a value in `0:ng-1` on all three axes and every index in range.

**Dropped — the pass-2 sweep in `_combine_kq_dedup_dense`.** It is provably in range (the loop bounds *are* the array dims) but it is O(prod(ngrid)) rather than O(pairs), and removing the annotation cost nothing measurable, so it is gone.

**Kept — the two device gathers in `run_eph_over_k_and_kq.jl`** (on `gpubte-opt`, #11, not this PR). Bounds-checking a device *index array* costs a Bool map+reduce kernel plus a D2H of the result per (k, tile); already justified in a comment there.

Related, and the reason I am confident about the one I kept: `_wrap_reduced` silently returned an out-of-range value for arguments outside `(-ng, 2ng)`, which under `@inbounds` is a segfault rather than a wrong answer — it killed the Julia process once during development. It now checks its range in a `@boundscheck`, so the `Dict` path and any future caller are covered, while the hot loop still elides it. That is the layering your comment is asking for: the annotation is only where it pays, and the precondition it depends on is checked once, cheaply, at entry.

PR #15, opened 2026-09-14
Move the (k, band) selection utilities from the kernel-ID prototype into band_states.jl

jaemolihm:
## Motivation

The Migdal-Eliashberg kernel-ID prototype (`superconductivity/dev_kernel_id_v3.jl`) accumulated a
handful of `(k, band)`-selection utilities that have nothing to do with Eliashberg theory — they
only manipulate a selection and the point group. They belong next to `unfold_band_states`, so the
prototype (and anything else that needs to restrict a sweep to a few states, or chase a k-point
across two grids) can use them instead of re-deriving them.

Two of them exist because the existing API does not cover the case:

* `filter_electron_states` rebuilds a selection from an energy window, so it cannot subset a
  *prebuilt* selection.
* `unfold_band_states` unfolds a whole selection into a **new** `FilteredBandStates` with its own
  k-grid, so it cannot answer "where do the images of this one state sit in *that* selection?".

## Change

**`refactor:` commit — pure rename, no behavior change.** `FilteredStates` → `FilteredBandStates`
(and `electron_states_to_FilteredStates` likewise). It and `BandStates` are the two subtypes of
`AbstractBandStates`; the lean one now reads as part of that family.

**`feat:` commit — `src/band_states.jl`.** Three routines promoted from the prototype, plus two
small API additions that fell out of them:

| added | does |
|---|---|
| `filter_states(s, keep)` | subset of a selection, **same concrete type** — so it works on `BandStates` (carrying `es`/`vs`) as well as `FilteredBandStates` |
| `state_index_in_star(s, xk, iband, symmetry)` | state index, searching the symmetry star of `xk` when the exact k-vector is absent |
| `state_indices_full_star(s, xk, iband, symmetry)` | *every* image of `(xk, iband)`, as indices into an existing selection |
| `getindex(::FilteredBandStates, i)` | the `(ik, iband, xk, weight)` NamedTuple — no `e`; both subtypes are now indexable and iterable |
| `state_index(s, st)` | item form, `st` anything carrying `xk`/`iband` |

Design decisions:

* **`filter_states` drops k-points left with no kept state**, because consumers iterate `kpts.n`
  (an outer-k e-ph loop visits every k-point of the grid). `get_filtered_kpoints` preserves
  `ngrid`, so the subset stays commensurate with the grid it was cut from. Per-state weights are
  carried over unchanged, so a subset's weights no longer sum to the full BZ.
* **Per-state item stays an isbits NamedTuple**, one `getindex` method per concrete type, rather
  than a new `BandStateItem` struct with a `e`-shaped hole in it. Both carry the `(xk, iband)`
  identity, which is all the item form of `state_index` needs.
* **`state_index(s, st)` goes through `xk`, not `ik`** — the two selections' k-grids are
  independent, so only the k-vector is a shared address. It turns

  ```julia
  state_index(dst, src.kpts.vectors[src.iks[i]], src.ibands[i])   # before
  state_index(dst, src[i])                                        # after
  ```

* The old `restrict_states` computed the old→new k-index map as `cumsum(mask)`, correct only
  because `get_filtered_kpoints` preserves order. Rewritten as an explicit scatter with that
  invariant named. Its docstring also claimed to be "the inverse of `unfold_band_states`", which
  it is not — it neither folds to the IBZ nor takes a whole selection.

## Test

New `test/test_band_states.jl` (51 assertions, model-free — states built by hand on a 4×4×4 grid
with the fcc point group), wired into `runtests.jl`:

* `getindex`/`iterate` — field sets per subtype, `eltype` matches, `isbitstype`, `collect` round-trip
* `state_index` item form — round-trips between a `FilteredBandStates` and a `BandStates` over the same grid
* `filter_states` — same concrete type, `(ik, iband)` sort order (required by `ind_range_for_k_range`),
  bare k-points dropped, `ngrid` preserved, weights/`nw`/`nstates_base` carried, `es`/`vs` subset,
  identity and empty subsets
* symmetry stars — full star vs. a brute-force `state_index` over every symop; exact-hit
  short-circuit; a k-point absent from the selection but with an image present; a band no image carries

Also checked, not committed: `filter_states` reproduces the prototype's `restrict_states`
bit-for-bit (every field, including `indmap` and `band_extent`) on 200 random subsets.

Existing suite: `test_filter_electron_states.jl`, `boltzmann/test_multigrid_weights.jl`,
`boltzmann/test_multigrid_transport.jl` — the tests that touch the renamed type — 301/301 pass.
The GPU arm of the multigrid test skipped (CUDA not functional on the machine I ran this on), so
that path is unverified here; the change is host-side only.


jaemolihm:
reopening to pick up the new base (fix(timer) landed on main)

src/band_states.jl
```diff
@@ -405,6 +419,81 @@ function state_index(s::AbstractBandStates{T, <:GridKpoints},
...
+
+# States `ks` of `s`, re-gridded onto `kpts_new` with the new per-state k-indices `iks_new`,
+# rebuilt as the same concrete type as `s`.
+_rebuild_states(s::FilteredBandStates, kpts_new, iks_new, ks) =
+    FilteredBandStates(kpts_new, iks_new, s.ibands[ks];
+        nw = s.nw, weights = state_weights(s)[ks], nstates_base = s.nstates_base)
+
+_rebuild_states(s::BandStates, kpts_new, iks_new, ks) =
+    BandStates(kpts_new, iks_new, s.ibands[ks], s.es[ks];
+        nw = s.nw, v = isempty(s.vs) ? eltype(s.vs)[] : s.vs[ks],
```
jaemolihm: 1. `nw = s.nw` -> `s.nw` in kwarg, same for nstates_base. record this to Julia developer agent
2. can isempty(s.vs) be  true?
3. why state_weights? we don't have direct access to weights?
4. 한국어로 왜 _rebuild_states 가 따로 필요한지 설명하세요

jaemolihm: All four addressed in 30d02f6.

**1. `nw = s.nw` → `s.nw`.** Done, both kwargs in both methods. Already the house spelling — `occ_fermion(e - μ, T; occ.occ_type)` in `src/common/occupation.jl:105`.

**2. Can `isempty(s.vs)` be true?** Yes — the public constructor defaults `v` to empty, and the field is documented "empty if not computed":

```julia
bs = BandStates(kpts, [1,2], [3,3], [0.1,0.2]; nw = 8)   # no v
isempty(bs.vs)                                            # true
```

Nothing inside the package reaches it (`electron_states_to_BandStates` always passes `v`), but the constructor allows it, and without the guard `s.vs[ks]` throws a `BoundsError`. Kept it, simplified to `isempty(s.vs) ? s.vs : s.vs[ks]`, and added a regression test.

**3. Why `state_weights`?** You're right, no reason — `state_weights(s) = s.weights` verbatim. Now `s.weights[ks]`.

**4. `_rebuild_states`가 왜 따로 필요한지**

`filter_states`의 앞부분 — 남길 상태를 `(iks, ibands)` 순으로 정렬하고, 남는 k점만 골라 `get_filtered_kpoints`로 새 k그리드를 만들고, 옛 k인덱스를 새 인덱스로 옮기는 부분 — 은 두 서브타입에 대해 **완전히 동일**합니다. 타입마다 달라지는 건 마지막 재구성 한 단계뿐입니다:

- `FilteredBandStates(kpts, iks, ibands; ...)`
- `BandStates(kpts, iks, ibands, es; ..., v)` — 에너지가 위치 인자이고, `es`/`vs`도 같이 잘라야 함

선택지는 셋이었습니다:

1. 타입별로 `filter_states`를 두 개 쓴다 → k그리드 재매핑 로직이 그대로 복붙됩니다.
2. `filter_states` 안에서 `if s isa BandStates`로 분기한다 → `FilteredBandStates`에는 존재하지도 않는 `s.es`/`s.vs` 필드 접근이 공용 메서드 본문에 들어갑니다. 죽은 분기라 실행은 안 되지만, 읽는 사람 입장에서 "이 타입에 저 필드가 있나?"를 매번 확인해야 하고 리팩터링에 취약합니다.
3. 공통 부분은 한 번만 쓰고, 다른 한 단계만 디스패치로 분리한다.

3번을 골랐습니다. 중복이 없고, 각 메서드의 반환 타입이 구성상 구체 타입으로 정해집니다. 이유를 코드 주석에도 적어 두었습니다.

한 가지 덧붙이면, 지금은 `BandStates` 메서드를 부르는 곳이 테스트밖에 없습니다 (v3은 `FilteredBandStates`만 자릅니다). 제네릭하게 가자고 결정하신 것이라 넣었는데, 좁혀서 `FilteredBandStates` 전용으로 되돌리는 것도 괜찮습니다 — 말씀만 주세요.

PR #18, opened 2026-09-14
Return (BandStates, imap) from both electron_states_to_BandStates methods

jaemolihmbot:
## Motivation

The two `electron_states_to_BandStates` methods returned different shapes:

| method | returned |
|---|---|
| `(el_states, kpts, nstates_base)` | `(BandStates, imap)` |
| `(el_states, sel::FilteredBandStates)` | `BandStates` |

So a caller switching from the k-grid method to the selection method — which is what
MigdalEliashberg.jl/#3 does, to stop losing the selection's per-state BZ weights — loses the
physical-band `imap` its loop indexes and has to rebuild it locally. That map belongs here, next to
`band_range`, `indmap` and `_indmap_to_device`, not in each downstream package.

## Change

`src/band_states.jl`

- Both methods now return `(BandStates, imap)`. Identical return type from both:
  `Tuple{BandStates{T,GridKpoints{T}}, OffsetMatrix{Int64,Matrix{Int64}}}`, so the two are
  interchangeable at a call site.
- The map is `OffsetArray(bs.indmap, band_range(bs), 1:kpts.n)`, inlined at both sites.
- The k-grid method no longer allocates and fills a second `imap` inside its flatten loop — the
  returned map is a view of the `BandStates`' own `indmap`, which `_build_indmap` already
  populates with the same data. Same rows (`band_range` is `iband_min:iband_max`), one fewer
  `nband × nk` allocation and one fewer store per state.

Callers updated to destructure: `boltzmann_calculator.jl` (`calc.el_i, _ = …`) and
`test/boltzmann/test_multigrid_transport.jl`.

## Test

| suite | pass | fail | error |
|---|---|---|---|
| `origin/main` @ `19808bd` | 22692 | 0 | 0 |
| this branch | 22692 | 0 | 0 |

Run on `workergpu068` (A100-80GB, CUDA functional). Behaviour-preserving, as expected — the map
contents and row offsets are unchanged; only who builds it moved.

Downstream: MigdalEliashberg.jl#3 depends on this and drops its local copy of the builder.


src/band_states.jl
```diff
@@ -115,6 +115,12 @@ function _build_indmap(n, nk, nband, nband_ignore, ik, iband)
     indmap
 end
 
+# `(physical band, ik) -> state index` view of a state set's `indmap`, whose rows are offset by
+# `nband_ignore`. 0 where the state is absent. Both `electron_states_to_BandStates` methods return
+# this, so a caller's loop indexes by physical band without knowing the offset.
+_physical_band_indmap(s::AbstractBandStates) =
```
jaemolihm: remove function, just inline in 2 callsites

jaemolihmbot: Done in `40c495e` — helper removed, inlined at both sites:

```julia
bs, OffsetArray(bs.indmap, band_range(bs), 1:kpts.n)      # kpts method
bs, OffsetArray(bs.indmap, band_range(bs), 1:sel.kpts.n)  # sel method
```

Suite on `workergpu068` (A100-80GB, CUDA functional): 22692 pass / 0 fail / 0 error, unchanged from `origin/main`.

Note the earlier hash `65bf04e` in this PR's description is dead — amended, not added to.

jaemolihm (approved):

PR #4, opened 2026-07-11
Stage 2b — shared GPU optimizations (consolidated dedup)

jaemolihm:
## Stage 2b — Shared GPU optimizations (consolidated dedup)

Fourth stage of the GPU-port merge (see `MERGE_PLAN.md`). Consolidates the three GPU
optimizations that were each implemented **twice** (identically) on `gpu-fmp-speedup` and
`bte-gpu`, so they are reviewed and merged exactly once. Cherry-picked from the `gpu-fmp-speedup`
copies onto the post-Stage-2a `main`.

### Commits
1. `perf(gpu): parallelize the fused e-ph rotation kernel over (band, band, q)` — applied clean.
2. `perf(gpu): k-side window projection of the batched e-ph interpolation` — **reconciled** (see below).
3. `perf(gpu): skip device-side bounds checks of the per-chunk iq gathers` — **reconciled**.
4. `docs: clarify get_eph_RR_to_kR_batched! uniform-nband requirement` — review follow-up.
5. `docs(bench): record Stage 2b GPU perf`.

### Reconciliation (the part to scrutinize)
Stage 2a had already (a) renamed the batch vars on `main` (`k_batch_size`→`nk_batch_max`,
`ks_tile`→`ks_batch`, `a`→`ik_ind`, `nqc`→`nq_batch`/`rng_q`, …) and (b) **removed the old
`batched`/host-fallback path** (the GPU loop now requires every calculator to implement the
batched device hook). The fmp commits predate both, so cherry-picks #2/#3 conflicted heavily.
Every conflict was resolved by keeping **main's structure and names** and layering in **only**
the window-projection + `@inbounds` semantics — no stale `kb`/`qb`/`ks_tile`/`nqc`/host-fallback
leaked back.

The k-side window projection lets the outer-k side carry only an `nbk_max`-wide contiguous
eigenvector window (not all `nw` bands), shrinking the kR→kq GEMM, both gauge rotations, `g2`,
and the scatter by `nw/nbk_max`. It adds a new `nbk0` band-offset kwarg to the
`run_calculator_batched!` contract; `nbk0 = 0` full-band runs are bit-identical to before.

### Verification
- **Critical review** (reviewer persona over the full diff): clean — verified the window-projection
  replay is byte-identical to fmp `15d2570` re-expressed in main's names, fullband path unchanged,
  `@inbounds` layered on main's `rng_q` gathers, `ndata` set-before-build ordering load-bearing.
[...]

PR #21, opened 2026-09-15
Add an item form of state_indices_full_star, matching state_index

jaemolihmbot:
## Motivation
`state_index` has an item form — `state_index(s, other[i])` — so a state of one selection is
located in another without unpacking it. `state_indices_full_star` did not, so every caller that
had a state in hand had to take it apart:

```julia
st = states_i[i]
append!(J, state_indices_full_star(states_f, st.xk, st.iband, symmetry))
```

Asked for in the MigdalEliashberg.jl PR #4 review, where the symmetry unfolding of the cross's
column pivots is exactly this loop.

## Change
`src/band_states.jl` — one method next to the existing one, delegating to it:

```julia
state_indices_full_star(s::AbstractBandStates, st, symmetry) =
    state_indices_full_star(s, st.xk, st.iband, symmetry)
```

* Takes anything carrying `xk` and `iband`, matching `state_index(s, st)`; the docstring says so
  in the same words.
* Purely additive: the 4-arg form is untouched, and the 3-arg signature cannot be ambiguous
  with it.

## Test
`test/test_band_states.jl`, in the existing "symmetry-star lookups" testset: the item form of a
state of `sel` returns the same `J` as the `(xk, iband)` form.

`BandStates: selection utilities` — 73/73 pass.


src/band_states.jl
```diff
@@ -484,12 +484,18 @@ end
...
 """
     state_indices_full_star(s, xk, iband, symmetry) -> Vector{Int}
+    state_indices_full_star(s, st, symmetry) -> Vector{Int}   # `st` a per-state item
 
 Indices in `s` of every image `(S·xk, iband)` of `(xk, iband)` under `symmetry`, sorted and
 deduplicated. Images absent from `s` (out of window, or at a k-point `s.kpts` does not hold) are
 dropped, so the result can be shorter than the group order and empty. Unlike `unfold_band_states`,
 which unfolds a whole selection into a NEW `FilteredBandStates` carrying its own k-grid, this
 returns indices into an EXISTING selection, for one state at a time.
+
```
jaemolihm, line 494: update the comment to just say you can pass `states[i]`

jaemolihmbot: Done in `c878830`:

```
The item form takes a state as `states[i]` yields it, like `state_index(s, st)`.
```

The signature block above already carries the `# `st` a per-state item` note, so that is all the paragraph needed to add.

PR #6, opened 2026-07-12
Stage 4 — GPU Boltzmann transport (BTE) calculator

jaemolihm:
Stage 4 of the GPU-port merge (see `MERGE_PLAN.md`). Adds the **GPU Boltzmann transport (BTE) calculator**, the transport analogue of the EliashbergCalculator, plus the shared `BandStates` type. Cherry-picked from `bte-gpu` onto the post-Stage-3 `main`, with the reconciliations below.

## Commits
- **`feat` GPU Boltzmann transport (BTE) calculator** (`481b310`) — new `GPUBoltzmannCalculator` (`src/boltzmann/gpu_boltzmann_calculator.jl`), `BandStates` (`src/boltzmann/band_states.jl`), device-safe shared physics core `bte_scattering_increments` (`src/boltzmann/bte_scattering_core.jl`), CUDA `bte_window_scatter!` kernel, `electron_lbte.jl` `BTorBandStates` union, example + regression test.
- **`perf` symmetry-reduce the k+q window filter** (`950ab35`) — filter on the IBZ + unfold the kept k-point set; composes with Stage 3's MPI outer-k path.
- **`feat` k-side projection offset** (`1fe846a`) — `run_calculator_batched!` accepts the batched loop's band offset.
- **`test` add CUDA to the test target** — `Pkg.test()` now runs GPU tests; test env derived entirely from `Project.toml` (fixes the recurring "IterativeSolvers not found" in ad-hoc envs).
- **`refactor` address critical review** — drop dead `ni_stride` param + vestigial `gpu_resident` knob; fix stale `BoltzmannCalculatorX2X`/`dev_BTE.jl` refs; soften "bit-identical"; example path via env var.
- **`fix` broaden the Method2-6 zero-δ NaN guard** — per-δ product guard (`0·Inf → NaN`); partial, with a physics TODO.
- **`bench` BTE benchmark + perf log**.

## Reconciliations onto current `main`
- `481b310` added `bte_window_scatter!` to the now-removed `src/wannier_to_bloch_gpu.jl`; placed the generic method in `src/calculator/calculator_utils.jl` next to `eph_window_scatter!`, included `bte_scattering_core.jl` in the boltzmann group.
- Calculator hooks renamed to Stage-2a API: `setup/flush_calculator_tile!` → `setup/flush_calculator_outer_batch!`.
- Offset kwarg renamed `nbk0` → `ibandk_offset` (0-based) to match the Stage-2b driver call.

[...]

PR #22, opened 2026-09-15
Add ElectronEigenpairs, a shareable full-band electron eigenpair cache

jaemolihmbot:
> **Stacked on #25.** The `xk_to_ik` / `xk_to_ik_unsafe` split that this PR's lookup depends on now lives in #25, based on `main`; merge that first. The commits here are unchanged in content, only rebased onto it.

## Motivation

Two separate e-ph runs over the same k-points do **not** share an eigenvector gauge. `g2 = |g|²` is gauge-invariant for a non-degenerate band, but inside a degenerate multiplet only the multiplet *sum* is — so two runs that split a multiplet differently disagree on `g2` at every state in it.

Today they agree only because both diagonalize a bitwise-identical `H(k)` with a deterministic LAPACK call. That is a coincidence, not a guarantee:

- `fourier_mode = "gridopt"` vs `"normal"` differ by `max|Δe| = 1.3e-15` in `H(k)` — and that flips a multiplet basis by **`max|Δu| = 1.57`**, i.e. O(1);
- the device path applies **no** degenerate-multiplet gauge fix (`compute_states.jl:131-134` vs `diagonalize.jl:49-84`); `max|u_gpu - u_cpu| = 1.72` on a plain 4³ Pb grid, where 22 of 64 k-points carry an exact degeneracy;
- a system with time-reversal **and** inversion is Kramers-degenerate at *every* k, making this pervasive rather than a corner case.

The gauge lives entirely in `u_full`, which is full-band and **window-independent**. So sharing the full-band eigenpairs is enough to pin the gauge across runs — no need to share whole `ElectronState`s — and each run still computes its own window-dependent `v`/`rbar`/`vdiag` from its own `quantities` list.

This PR is the cache only. **Nothing consumes it yet**; `compute_electron_states` and the driver keywords follow.

## Why a new type

The gauge lives in `ElectronState.u_full` (`electron_state.jl:33`). Nothing else in the package carries it, and the one thing that does carries too much:

| candidate | why not |
|---|---|
| `FilteredBandStates` | **Pre-eigenvector by design** — "Lean selection of `(k-point, band)` states, **before eigenvectors/velocities are computed**" (`band_states.jl:91-94`). It has no `u` at all, and `compute_electron_states(model, sel, …)` re-diagonalizes from it, so passing one pins nothing. |
| `BandStates` | Adds only `es`/`vs` to that selection. Still no eigenvectors. |
| `Vector{ElectronState}` | Does carry `u_full` — but everything *else* in it (`v`, `rbar`, `vdiag`, `occupation`) is a window-restricted `OffsetArray` view keyed on `rng` (`electron_state.jl:68-81`), so sharing it couples the window across runs. It also records **no list of which quantities were filled**, so a vector built with fewer quantities silently hands back uninitialized `v`/`rbar`. And it is mutable and mutated in place — `set_occupation!(epstate.el_k, …)` writes through the `epstate.el_k = el_k_save[ik]` alias in several calculators. |

Combining them does not help either: the two `*BandStates` contribute no gauge, so any combination reduces to `Vector{ElectronState}` plus a selection, and inherits its window coupling and its missing quantity record.

The gauge itself is **full-band and window-independent**. So the minimal sufficient object is `(e, u)` over a k-grid — which is what this type is. A partially-filled state is unrepresentable: holding `e` and `u` *is* the guarantee, and each run still computes its own window-dependent quantities from its own `quantities` list.

## Change

`src/electron_eigenpairs.jl` (new, +111):

- **`ElectronEigenpairs{T, MT<:AbstractMatrix{T}, AT<:AbstractArray{Complex{T},3}}`** — `nw`, `kpts::GridKpoints{T}`, `e (nw, nk)`, `u (nw, nw, nk)`. Immutable, no mutating API; the inner constructor checks the three shapes. **`e`/`u` follow the backend** (same parameter shape as `WannierObject`'s `op_r` and `KRtoKQWorkspace`), so a GPU-built cache stays on the GPU and is consumed without a round trip; `kpts` and the lookup stay host-side. `u` is `16·nw²·nk` bytes resident for the cache's lifetime — modest for a windowed selection, large for a dense full-BZ grid.
- **`electron_eigenpairs(model, kpts; fourier_mode = "gridopt", backend = CPUBackend())`** — CPU: the `@threads`-over-chunks + thread-local interpolator pattern of `_compute_electron_states_cpu!`, calling `get_el_eigen!` into column views. Device: one `get_el_eigen_batched`, copied straight back to the host — no device-resident state.
- **`_eigenpair_index(eig, xk)`** — O(1) `xk -> ik` on `_hash_xk` / `_ik_from_hash`.

Two design points worth calling out:

| decision | why |
|---|---|
| `kpts` is concretely `GridKpoints`, not `<:AbstractKpoints` | the lookup is grid-only (`ngrid`, `shift`), so a `Kpoints`-backed cache is **unconstructible** rather than a `MethodError` at lookup |
| the lookup validates the **query**, not only the cached points | `_hash_xk` rounds, so an off-grid `xk` would otherwise alias to a neighbouring node and silently return the wrong k's eigenvectors. Same criterion and `atol` as `GridKpoints(::Kpoints, ngrid)`, so no k-list that can become a `GridKpoints` is newly rejected |

`atol` applies in **grid-index** units, so the accepted window is a fixed `1.5e-8` grid spacings at any `ngrid` and cannot reach a neighbour. Measured residuals: `kpoints_grid` is bit-exact to 1000³; `combine_kpoint_grids` at 144³ sits 3.0e5x below the tolerance.

## Test / Usage

```julia
eig = electron_eigenpairs(model, kpts)                  # or backend = gpu_backend()
ik  = ElectronPhonon._eigenpair_index(eig, xk)
eig.u[:, :, ik]                                         # the full-band gauge at that k
```

`test/test_electron_eigenpairs.jl` (new, +113):

| arm | assertion |
|---|---|
| CPU vs `compute_electron_states` | **bitwise** `==` for `e` and `u`, 64/64 k, both `fourier_mode`s |
| negative control | no k's `u` is bitwise equal to its neighbour's, so the above cannot go vacuous |
| lookup | own points resolve; `+1e-12` round-off tolerated; off-grid and uncovered-node rejected with **distinct** messages; a shifted cache grid rejects Γ |
| GPU vs CPU | eigenvalues `7.4e-16` relative, unitarity `3.0e-15`; `u` reconstruction **partitioned by degeneracy** — non-degenerate `< 1e-13` (4.6e-15), degenerate `< electron_degen_cutoff` (2.2e-8) |
| eltype | `to_device(gpu_backend(), zeros(ComplexF64, 2, 2)) isa CuArray{ComplexF64}` — `false` for `cu`, which demotes to `ComplexF32` |

The GPU arm compares **eigenvalues, not eigenvectors**: the device path has no multiplet gauge fix, so `u` may legitimately differ inside a degenerate subspace. The `u` check is on `u·diag(e)·u'` instead, partitioned so the 42 non-degenerate k-points are held to round-off rather than to the degenerate cutoff.

- **37/37** with CUDA (7 of them the GPU arm), host `workergpu068` / A100-80GB.
- **30/30** in a second env with `CUDA` removed from `[deps]` — `CuArray` is not even resolvable there, so the 37-30 = 7 delta proves the GPU arms skip cleanly and the base package still loads without its extension.
- Regression sweep unchanged: `test_ElectronState` 23/23, `test_kpoints` 16602/16602, `test_diagonalize` 10/10, `test_velocity` 65/65.


src/electron_eigenpairs.jl
```diff
@@ -0,0 +1,111 @@
...
+                                u::Array{Complex{T}, 3}) where {T}
+        size(e) == (nw, kpts.n) || throw(ArgumentError(
+            "e must be of size ($nw, $(kpts.n)), got $(size(e))"))
+        size(u) == (nw, nw, kpts.n) || throw(ArgumentError(
+            "u must be of size ($nw, $nw, $(kpts.n)), got $(size(u))"))
+        new{T}(nw, kpts, e, u)
+    end
+end
+
+nk(eig::ElectronEigenpairs) = eig.kpts.n
```
jaemolihm: nk is too generic name to use as a function. why not just eig.kpts.n

src/electron_eigenpairs.jl
```diff
@@ -0,0 +1,111 @@
...
+
+The value is immutable and read-only; there is no mutating API. Look a k-point up by its crystal
+coordinates with `_eigenpair_index`.
+"""
+struct ElectronEigenpairs{T}
+    nw   :: Int
+    # `GridKpoints` rather than `AbstractKpoints`: the xk -> ik lookup is defined only for a grid
+    # (`_hash_xk`), so requiring one here is what makes every cache lookupable.
+    kpts :: GridKpoints{T}
+    e    :: Matrix{T}               # (nw, kpts.n)
```
jaemolihm: isn't it better for e and u to live on device ?

src/electron_eigenpairs.jl
```diff
@@ -0,0 +1,111 @@
+# Full-band electron eigenpairs over a k-point set, computed once and shared by several e-ph runs.
```
jaemolihm: add "to ensure eigenvector gauge consistency"

src/electron_eigenpairs.jl
```diff
@@ -0,0 +1,111 @@
...
+inside a multiplet; and the whole set is one batch, so the device H(k)/U stacks (nw^2 * nk) are
+unbounded and a very large k-grid can OOM.
+"""
+function electron_eigenpairs(model::Model{FT}, kpts; fourier_mode = "gridopt",
+                             backend = CPUBackend()) where {FT}
+    (; nw) = model
+    gkpts = GridKpoints(kpts)
+    e = zeros(FT, nw, gkpts.n)
+    u = zeros(Complex{FT}, nw, nw, gkpts.n)
+    if backend isa CPUBackend
```
jaemolihm, line 97: put `@views` here and use ` view(gkpts.vectors, iks)` -> `gkpts.vectors[iks]`

jaemolihm:
Add why we need new type, why combining BandStates/FilteredBandStates and Vector{ElectronStates} will not work

src/electron_eigenpairs.jl
jaemolihmbot: Dropped in `6523f49`. Callers use `eig.kpts.n`; `isdefined(ElectronPhonon, :nk)` is now `false`, and the Stage 1 and Stage 2 uses are gone.

The reviewer had independently flagged the same thing from the other side: `nk` is a local variable in 20+ `src/` functions (`compute_states.jl:166`, `run_transport.jl:127`, …), so a module-scope `nk` made the name mean two things in one namespace. It suggested keeping the accessor; your version is better — the `size(eig.e) == (nw, kpts.n)` assertions already pin the quantity, so nothing was lost.

src/electron_eigenpairs.jl
jaemolihmbot: Done in `6523f49`. The header line now reads:

> … shared by several e-ph runs to ensure eigenvector gauge consistency.

which is the first thing a reader needs, since the type is otherwise just two arrays and a k-grid.

src/electron_eigenpairs.jl
jaemolihmbot: Done in `6523f49` — the thread-chunk block is `@views` and the expression is `gkpts.vectors[iks]`.

Checked the semantics rather than assuming: `@views` over that expression and array type yields a `SubArray`, and the macro is *inside* the `@threads` body, so there is no macro-ordering question about it wrapping the loop construct.

src/electron_eigenpairs.jl
jaemolihmbot: Done in `6523f49` — `ElectronEigenpairs{T, MT<:AbstractMatrix{T}, AT<:AbstractArray{Complex{T},3}}`, so `e`/`u` follow the backend. The parameter shape is copied from `WannierObject`'s `op_r` and `KRtoKQWorkspace{MT, AT}` rather than invented.

You corrected my framing when I raised this, and it matters enough to record here: **the hazard is two runs on the *same* backend diverging through floating point**, not sharing a cache across backends. I had argued for host-residency on the grounds that a host cache is the common denominator between a CPU and a GPU run — that was never the requirement.

What follows from your framing:

- `electron_eigenpairs(model, kpts; backend)` returns arrays resident on `backend`; the device builder keeps `get_el_eigen_batched`'s output where it produced it. A GPU run's cache is consumed with no round trip (#23).
- `kpts` and `_eigenpair_index` stay **host-side** — only `e`/`u` move. The lookup is O(1) host arithmetic and belongs nowhere else.
- Cache and consumer must agree: `_check_eigenpairs(eigenpairs, nw, backend)` in #23 throws an `ArgumentError` naming both sides, rather than a silent host copy or a scalar-indexing error deep in a loop.

One correction to something I said earlier in this thread's discussion, so it is not left standing: I described this as reversing a "recorded 2026-06-20 decision". Having read the source, that decision (`GPU_PROGRESS.md`, DEVICE-VELOCITY) is about `ElectronState`/`PhononState`, predates this type, and its own revisit clause — "revisit only if a GPU transport calculator appears" — has already fired (`GPUBoltzmannCalculator`, PR #6, and ME's `EliashbergCalculator`; see `plans/batched_phonon_state.md:37-50`). So this is a new type outside that decision's scope, not an exception to it.

The one premise from it that **does** still apply is kept as a fact in the docstring rather than a decision: `u` is `16·nw²·nk` bytes resident for the cache's lifetime — modest for a windowed selection, large for a dense full-BZ grid.

src/electron_eigenpairs.jl
```diff
@@ -0,0 +1,121 @@
...
+
+On a GPU backend the whole set is solved in one batched eigensolve and `e`/`u` are left on the
+device, so the cache is resident on `backend` and a consumer must run on that same backend.
+`fourier_mode` is then unused (the batched interpolator is the only one the device path has), as in
+`compute_electron_states`. Two caveats inherited from that path: the batched eigensolve does not
+apply the degenerate-multiplet gauge fix of the per-k solve, so a device-built cache is
+self-consistent but differs from a CPU-built one inside a multiplet; and the whole set is one batch,
+so the device H(k)/U stacks (nw^2 * nk) are unbounded and a very large k-grid can OOM.
+"""
+function electron_eigenpairs(model::Model{FT}, kpts; fourier_mode = "gridopt",
```
jaemolihm, line 93: seems overlapping with compute_electron_states. Any way to reduce redundancy ?

src/electron_eigenpairs.jl
```diff
@@ -0,0 +1,121 @@
...
+        E_dev, U_dev = get_el_eigen_batched(itp_elham, gkpts.vectors)
+        ElectronEigenpairs(nw, gkpts, E_dev, U_dev)
+    end
+end
+
+# Index of `xk` in the cache, erroring on a query the cache cannot answer. Both failures are loud:
+# `_hash_xk` rounds `xk` onto the cache's grid, so an off-grid query would otherwise alias to the
+# nearest cached node, and `_ik_from_hash` returns 0 for a node the cache does not hold -- neither
+# may reach a consumer as an index.
+function _eigenpair_index(eig::ElectronEigenpairs{T}, xk) where {T}
```
jaemolihm: isn't this just xk_to_ik(eig.kpts, xk) ?

src/electron_eigenpairs.jl
```diff
@@ -0,0 +1,121 @@
...
+eigensolve does not pin it at all.
+
+`e` and `u` are `AbstractArray`s that live wherever the backend that built them put them, as the
+`op_r` of a `WannierObject` does: a cache built with `backend = gpu_backend()` stays on the device
+and is consumed there without a round trip. The runs that must agree on a gauge are runs on the
+*same* backend, so moving the arrays to the host would only mean uploading them again per run;
+a consumer on a different backend than the cache is an error, not a silent copy. `kpts`, and the
+`xk -> ik` lookup over it, always stay on the host.
+
+The size to keep in mind is `u`: `16 * nw^2 * nk` bytes, resident for the cache's whole lifetime.
```
jaemolihm: add In the future, filtering can be combined to store a subset of eigenvectors

In this regards, maybe rename u -> u_full (consistent with ElectronStates) ?

src/electron_eigenpairs.jl
jaemolihmbot: Yes — fixed in 32154e4.

`_eigenpair_index` now calls `xk_to_ik(xk, kpts)` (note the argument order is `(xk, kpts)`, `src/common/kpoints.jl:661`) and keeps only the two loud failures it adds on top:

* the on-grid validation of the *query* point — `_hash_xk` rounds `xk` onto the cache's grid, so without it an off-grid query aliases onto the nearest cached node instead of missing (this was review round 1's finding 1);
* the uncovered-node miss, i.e. `xk_to_ik`'s `nothing`.

`xk_to_ik`'s `Union{Nothing, Int}` does not leak into the callers: inference narrows past the `=== nothing` throw, so `Base.return_types(_eigenpair_index, ...) == [Int64]`, and in a probe of PR #23's index-then-view call site shape the `ik` slot is `Int64` with 0 `Core.Box` and 0 `Any` slots (`map`ping it gives `Vector{Int64}`).

Tallies unchanged at 32154e4: **38/38** with CUDA (GPU testset ran, 10 assertions) and **28/28** in a CUDA-absent env; 38 - 28 = 10. Host `workergpu068` (A100-80GB), `realpath(pathof(ElectronPhonon))` checked against the worktree before each tally.

src/electron_eigenpairs.jl
jaemolihmbot: Both done in 32154e4.

* The docstring's size paragraph now ends with the forward-looking note: *"In the future the filtering can be combined with the cache, storing only the subset of the eigenvectors that the runs actually need."*
* `u` -> `u_full`, matching `ElectronState.u_full` and saying full-band rather than windowed. Updated: the field, the inner constructor's argument, size check and error text, both builders, all three docstrings and the Stage 1 tests. `Base.show` prints only `nw`/`nk`/`typeof(e)`, so it needed no change.

Scope note: the rename is confined to this branch's two files. `src/compute_states.jl` here is byte-identical to `main`, so the `eigenpairs.u` uses in the Stage 2 consumer (#23) are not touched by this commit and will be fixed when that branch is restacked onto this tip.

Tallies unchanged at 32154e4: **38/38** with CUDA (GPU testset ran, 10 assertions) and **28/28** in a CUDA-absent env; 38 - 28 = 10.

src/electron_eigenpairs.jl
jaemolihmbot: Investigated, **no source change made** — leaving this open for you to decide (per your call: report options, don't patch).

## Every site that solves the electron eigenproblem over a k-set

| lines | site | what it is | overlap with the cache builder |
|---|---|---|---|
| 46 | `_compute_electron_states_cpu!` (`compute_states.jl:77-122`) | per-k threaded solve + window + velocity/position into `ElectronState`s | **6 lines**: the `@threads`/`chunks` preamble, `get_interpolator(model.el_ham; fourier_mode)`, `register_kpoints!`, the `for ik in iks` header, the solve call |
| 44 | `_compute_electron_states_device!` (`:135-178`) | one all-nk batched solve + batched velocity/position | **2 lines**: the `"batched"` `get_interpolator` and `get_el_eigen_batched` |
| 34 | `_scatter_electron_states!` (`:182-215`) | host copy of dense `E`/`U`/`rbar`/`vel` into the states | 0 — the builder has nothing to scatter into |
| 14 | `electron_eigenpairs` CPU branch (`electron_eigenpairs.jl:86-99`) | the cache's own solve | — |
| 6 | `electron_eigenpairs` device branch (`:100-105`) | the cache's own batched solve | — |
| 27 | `_filter_kpoints` device branch (`filter.jl:46-72`) | **chunked** batched value-only solve + window test | 2 lines, but chunked to a ~1 GiB stack |
| 18 | `_filter_kpoints` CPU branch (`filter.jl:74-91`) | per-k threaded value-only solve + window test | 4 lines |
| 25 | `filter_qpoints` (`filter.jl:109-133`) | per-q solve over all k+q with an early `break` | 0 — not a k-set pass |
| 12 | `compute_eigenvalues_el` (`compute_eigenvalues.jl:8-19`) | serial value-only solve, one scratch `ElectronState`, gathers `e_full` into a dense `(nw, nk)` | **this is already candidate (a)'s shape, in the tree today** |
| 1 each | `set_eigen!(epstate.el_kq, ham, xkq)` (`run_eph_over_k_and_q.jl:303`, `run_eph_over_q_and_k.jl:353`, `run_electron_phonon.jl:162`) | per-(k,q) inside the e-ph loop | 0 |

So the real overlap is 6 lines on CPU and 2 on device, not 46 + 44.

## Candidate (a) — the CPU builder calls `compute_electron_states(model, kpts, ["eigenvalue","eigenvector"])` and gathers

Measured (Pb artifact model, `nw = 4`, 16 threads, `fourier_mode = "gridopt"` on both arms, min of 2 timed reps after a warm-up):

* **Correct**: bitwise identical, `e == e` and `u_full == u_full` at every k, at nk = 13824 and nk = 32768.
* **Cost**, nk = 32768: **112 ms / 9.7 MiB** direct vs **224 ms / 102.9 MiB** via states — **2.0x** time, **10.6x** allocations. The extra is the transient `Vector{ElectronState}`: ~93 MiB, i.e. ~2.9 KiB of scratch per k point to carry a 4-value + 4x4 eigenpair, because each state also allocates `vdiag`/`v`/`rbar`/`occupation` at `nband_bound = 1`.
* **Lines**: the CPU branch goes 13 -> 8.
* **The device path cannot do the same, and it is a type that forbids it, not a preference.** `ElectronState`'s fields are `e_full :: Vector{Float64}` and `u_full :: Matrix{ComplexF64}` — concrete host arrays — and `_compute_electron_states_device!` never returns `E_dev`/`U_dev`; it ends in `_scatter_electron_states!(..., Array(E_dev), Array(U_dev), ...)`, so the only `e`/`u` a caller can reach are host copies. A device cache assembled from `ElectronState`s is therefore a full round trip: measured **6.4 ms -> 91.7 ms** at nk = 32768 (**14.4x**), 9.0 MiB down and 9.0 MiB back up — i.e. it gives back exactly the residency 6523f49 was asked to establish.

Net: (a) saves 5 lines in one branch, costs a layer and a 10x transient, and leaves the function asymmetric (CPU delegates, device must not).

## Candidate (b) — extract a shared `_eigensolve_all!`

Prototyped on the *most favourable* quantity list: pass 1 = the shared kernel writing `(nw, nk)` / `(nw, nw, nk)`, pass 2 = the existing `_scatter_electron_states!`.

* Results identical to today's one-pass (`e_full`, `u_full` and `rng` all `==` at every k).
* Cost, nk = 32768: **175 ms / 93.9 MiB -> 195 ms / 101.9 MiB** — +11% time, +9% allocations, the +8.0 MiB being the forced dense `(nw, nw, nk)` buffer that the states path does not materialize today.
* **Does redundancy actually go away? The 6-line preamble does, and nothing else.** For any quantity list containing velocity or position, `_scatter_electron_states!` is unusable (it wants dense `rbar`/`vel`), so the states worker needs a **second `@threads` region** whose `pos`/`vel` interpolators are registered on its own chunks — today all three are registered once on the same chunk. And `quantities == ["eigenvalue"]` runs `solve_eigen_el_valueonly!`, a different LAPACK driver, so it cannot share an eigenvector-producing kernel at all.

Net: -6 lines in one place, +1 forced buffer, +1 pass, +1 indirection for every `compute_electron_states` caller.

## Candidate (c) — keep both, document why

Cost 0. `compute_eigenvalues_el` is the standing precedent: the same "solve over a k-set, gather into a dense `(nw, nk)`" shape, 12 lines, coexisting with `compute_electron_states` today — `electron_eigenpairs`' CPU branch is its eigenvector-carrying, threaded twin. What is missing is the *why*: the type docstring explains the gauge purpose but not why the builder does not call `compute_electron_states`.

## 10% rule

The whole eigensolve pass over the k-set is **0.11 s at nk = 32768**, against an e-ph sweep the plan measures in millions of (k, q) pairs (plan, "Does it earn its keep on performance alone? No." — expected ≪ 1%). So **none** of the three is chosen on runtime; all are well under 10%, and the rule says take the fewest lines and fewest layers.

| candidate | lines | layers | device residency | runtime | verdict |
|---|---|---|---|---|---|
| (a) delegate to `compute_electron_states` | -5 (CPU branch only) | +1 | **lost** (14.4x, 2x9 MiB round trip) | 2.0x CPU build | rejected on the device branch alone |
| (b) shared `_eigensolve_all!` | -6 | +1, and +1 pass / +1 region for velocity lists | kept | +11% on `compute_electron_states` | buys 6 lines, taxes every caller |
| (c) keep both + a "why not reuse" comment | +3 comment lines | 0 | kept | 0 | **recommended** |

**Recommendation: (c)** — add the comment your rule asks for (why `compute_electron_states` cannot be reused: `ElectronState` is host-typed and per-k, the cache is backend-resident and whole-set), and leave the 20-line builder alone.

One aside, in case a dedup that *pays* is wanted: it is not on this list. `_filter_kpoints`' device branch is the only one that **chunks** the device H(k) stack (`kchunk` capped at ~1 GiB); both `_compute_electron_states_device!` (its own `TODO` at `compute_states.jl:128-130`) and `electron_eigenpairs`' device branch solve all nk in one batch and can OOM. That is behaviour *missing* in two places rather than duplicated in two places, and sharing it would be a real fix rather than a line count.

jaemolihmbot: Resolved in 5fa134e — you picked **(c)**, keep both paths and document why, so this is settled rather than open.

`electron_eigenpairs` now carries the "why not reuse" comment your rule asks for, stating the facts rather than the history:

* `compute_electron_states` returns one `ElectronState` per k point, whose `e_full`/`u_full` are host-concrete `Vector`/`Matrix`;
* its device worker has no device-returning exit — `_scatter_electron_states!` takes `Array(E_dev)`/`Array(U_dev)`, so `E_dev`/`U_dev` never leave the function — hence delegating would give up the backend residency this type exists to provide (measured on a 4-band model at nk = 32768: **2.0x** runtime and **10.6x** allocations on CPU, and a **14.4x** host round trip on device);
* it also interleaves the energy window and the derived quantities into the same thread-local pass, which a full-band cache has no use for;
* `compute_eigenvalues_el` (`compute_eigenvalues.jl:8-19`) is the existing precedent for a dense-gather twin coexisting with `compute_electron_states`.

The comment sits *above* the docstring, not between it and the function — a comment in that position detaches the docstring (verified: `@doc electron_eigenpairs` returned the no-docs fallback with it there, and the real docstring with it moved).

Also in this commit, per your other decision: **`e` -> `e_full`**, so the fields are now `e_full`/`u_full`, matching `ElectronState` exactly. `src/compute_states.jl` is still byte-identical to `main`.

Tallies at 5fa134e, host `workergpu068`: **38/38** with CUDA (GPU testset ran) and **28/28** CUDA-absent; 38 - 28 = 10 GPU-gated assertions, unchanged. Regression sweep unchanged: `test_ElectronState` 23/23, `test_kpoints` 16602/16602, `test_diagonalize` 10/10, `test_velocity` 65/65.

Noted on the device-batch OOM aside: no code change, and your reasoning holds — `_filter_kpoints` chunks because it sweeps the full grid, while this builder and `compute_electron_states` run over an already-filtered selection, and `run_eph` already holds a device `u_full` stack for the same k-set. The docstring's existing "the whole set is one batch ... a very large k-grid can OOM" caveat stays as the record.

src/electron_eigenpairs.jl
```diff
@@ -0,0 +1,129 @@
...
+        E_dev, U_dev = get_el_eigen_batched(itp_elham, gkpts.vectors)
+        ElectronEigenpairs(nw, gkpts, E_dev, U_dev)
+    end
+end
+
+# The cache's xk -> ik accessor. `xk_to_ik` reports a node the cache does not hold as `nothing`,
+# leaving the caller to say what a miss means, which here is always a bug: an index is the only
+# thing a consumer can do with the answer. Off-grid queries the checked lookup rejects itself --
+# the unchecked one would alias them onto a neighbouring cached node.
+function _eigenpair_ik(eig::ElectronEigenpairs, xk)
```
jaemolihm: do we really need this new function ? Can't we just use `xk_to_ik(xk, eig.kpts)` ? I don't want more abstraction...

jaemolihmbot: Agreed, and dropped — `fb7aa05` → `28d82f3`. The lookup is `xk_to_ik(xk, eig.kpts)` again, exactly as you wrote it.

What settles it is that the function had **no production caller in this PR at all**: its only uses here were the test file and two docstring lines, because the four real call sites live in #23's `compute_states.jl`. So in #22 it was pure surface.

Why it appeared in the first place, since it was not in the version you reviewed earlier: this PR was restacked onto the squash-merged #25, and the two commits it used to carry were a *pre-review* copy of #25 in which `xk_to_ik` threw for a node the set does not hold. What actually merged returns `nothing` there and leaves the caller to raise its own error — so the accessor was that error put back in one place. Dropping it here is right; the miss handling belongs to #23, and is being done at the call sites there per your steer.

One finding while doing it, for the record rather than for this PR. With no guard at all, a k point the cache does not hold reaches the gather as a `nothing` index, and the two backends fail very differently:

* CPU — `ArgumentError: invalid index: nothing of type Nothing`, wrapped by the threaded state loop in a `TaskFailedException`.
* GPU — the index is consumed *inside* the indexing kernel: `KernelException: exception thrown during kernel execution on device NVIDIA A100-SXM4-80GB`. No k point, no cache, not even `Nothing`.

So #23 checks for `nothing` at each call site, on the host, before the index reaches the gather. That is invisible to a CPU-only run, which is why it is worth stating here.

The test file keeps only the cache's own contract, as its commit message says — that a lookup addresses the column of `e_full`/`u_full` holding that k point's eigenpair, and that a cache whose grid does not contain its own points is unconstructible. `xk_to_ik`'s own behaviour stays covered by "kpoints: xk_to_ik checked vs unsafe" in `test_kpoints.jl`.

Suite at `28d82f3`, host `workergpu068` (A100, `CUDA.functional() = true`): **22743/22743**.

PR #23, opened 2026-09-15
Consume an ElectronEigenpairs cache in compute_electron_states

jaemolihmbot:
> **Stacked on #22** (`ElectronEigenpairs`, the cache type). Review that first — this PR's diff is Stage 2 alone. Merge #22 first; GitHub retargets this one to `main` automatically.

## Motivation

#22 added `ElectronEigenpairs` but nothing consumed it. This wires it into `compute_electron_states`, so a caller can hand several runs one set of eigenpairs and get **the same eigenvector gauge in each**.

That is the point of the feature: `g2 = |g|²` is gauge-invariant per band only when the band is non-degenerate, so two runs that split a degenerate multiplet differently disagree on `g2` at every state in it. The runs do not have to differ in any interesting way to diverge — a `1.3e-15` change in `H(k)` is enough to flip a multiplet's basis by O(1).

Without a cache nothing changes: the no-cache path is bitwise identical to `main`, and not slower.

## Change

- `eigenpairs = nothing` on both `compute_electron_states` methods, forwarded to both workers as a **typed positional** argument.
- `_set_eigen_from!` / `_set_eigen_valueonly_from!`, two methods each — `::Nothing` is today's `set_eigen!`, `::ElectronEigenpairs` is an `_eigenpair_index` lookup plus a column copy. **Dispatch, not a branch**, so `nothing` costs nothing in the per-k loop.
- Neither worker builds its `el_ham` interpolator when a cache is supplied; nothing else uses it.
- Device path slices the cache in place (`eigenpairs.u[:, :, iks]`) instead of calling `get_el_eigen_batched`. A `CuArray` indexed by a host `Vector{Int}` gathers correctly under `allowscalar(false)`, so the expression is identical on both backends — no `to_device(iks)` and no branch.
- **Backend-residency contract:** `_check_eigenpairs(eigenpairs, nw, backend)` runs once at the entry of both methods — the only place a cache and a run's backend meet. A mismatch is an `ArgumentError` naming both sides (*"eigenpairs are resident on the host (e::Matrix{Float64}), but the run uses GPUBackend; build the cache with the run's backend"*), rather than a silent host copy or a scalar-indexing error deep in a loop. An `nw` mismatch likewise fails at entry rather than as a `DimensionMismatch` inside a thread.

## Test / Usage

```julia
eig    = electron_eigenpairs(model, kpts)                      # or backend = gpu_backend()
states = compute_electron_states(model, kpts, quantities, window; eigenpairs = eig)
```

| property | evidence |
|---|---|
| no-cache path **bitwise** unchanged | **32/32** configurations identical under `==` — 2 `el_velocity_mode`s × 4 quantity lists × both methods × CPU/GPU — over `xk`, `e_full`, `u_full`, `nband`, `rng`, `vdiag`, `v`, `rbar`. Snapshotted on the base commit (whose `compute_states.jl` is byte-identical to `main`) and replayed on the branch. |
| no-cache path not slower | cubicBN, nk = 4096, min of 5: full quantity list **0.0646 s** vs **0.0672 s** baseline; value-only **0.0367 s** vs **0.0390 s**. With a cache, 0.0259 s. |
| cache path correct | **92/92** across the eigenpairs testsets with CUDA, **64/64** in an environment with `CUDA` removed from `[deps]` — the 28 difference is exactly the GPU arms, so they skip cleanly rather than vanish. |
| teeth | wrong-`nw` cache → `ArgumentError`; backend mismatch → `ArgumentError`, both directions; a cache missing one k point → the lookup's miss error, not a silent recompute. |

Regression sweep unchanged: `test_ElectronState` 23/23, `test_velocity` 65/65, `test_diagonalize` 10/10, `test_filter_electron_states` 15/15, `test_cubicBN_eigenvalues` 8/8, `test_cubicBN_spectral` 16/16, `test_cubicBN_selfen` 24/24, `test_gpu.jl` all testsets (CPU-vs-GPU rdiff 3.04e-16 / 8.11e-16).

### Two things that cannot be bitwise inert, by construction

Both are asserted on what they must do instead, and documented:

- **`fourier_mode`** — with a cache it no longer affects `e`/`u`, so inertness requires the cache to have been built with the same mode. Crossing `"gridopt"` and `"normal"` gives `max|Δu| = 1.57` from that `1.3e-15` difference in `H(k)`.
- **`quantities == ["eigenvalue"]`** — without a cache that branch runs the *value-only* LAPACK driver; a cache necessarily holds the *full* eigensolve's eigenvalues. EP's own two paths already disagree at `max|Δe| = 1.55e-15` with no cache involved.


src/compute_states.jl
```diff
@@ -74,14 +91,66 @@ function _electron_state_needs(model, quantities)
...
+_check_eigenpairs(::Nothing, nw, backend) = nothing
+
+function _check_eigenpairs(eig::ElectronEigenpairs, nw, backend)
+    eig.nw == nw || throw(ArgumentError(
+        "eigenpairs holds nw = $(eig.nw) Wannier functions, but the model has nw = $nw"))
+    # A cache is resident on the backend that built it, and this run's arrays live on `backend`.
+    # Consuming one from the other side would mean moving it per call, or indexing a device array
+    # from the host loop, so require a match.
+    on_host = eig.e_full isa Array
+    on_host == (backend isa CPUBackend) || throw(ArgumentError(
```
jaemolihm: do we need this? can copyto! implicitly do host -> device ? Or this is too costly so an error is better ?

src/compute_states.jl
```diff
@@ -74,14 +91,66 @@ function _electron_state_needs(model, quantities)
     (; need_vfull, need_vdiag, need_position, need_velocity)
 end
 
-function _compute_electron_states_cpu!(states, model::Model{FT}, kpts, quantities, window;
-                                       fourier_mode) where FT
+_check_eigenpairs(::Nothing, nw, backend) = nothing
+
+function _check_eigenpairs(eig::ElectronEigenpairs, nw, backend)
```
jaemolihm: I am not sure this safeguard is useful, without this one still will get an error, right ?

src/compute_states.jl
jaemolihmbot: Kept, but moved and generalized in `710e707`.

`copyto!` would not save it, in either direction:

* **host cache, device run.** The `quantities == ["eigenvalue"]` path does `Array(eigenpairs.e_full[:, iks])`, which on a host array just *works* and gives correct results — so that direction is not an error at all, it silently behaves differently from every other path. The full path instead feeds a host `U_dev` into the device batched GEMMs and fails somewhere unrelated.
* **device cache, CPU run.** `_set_eigen_from!` runs inside the `@threads` loop, once per k point, so an implicit copy is a device→host transfer per k — silently pathological rather than wrong.

So it is neither "copyto! handles it" nor purely "too costly": one direction has no error to fall back on. What changed is that the check is no longer written out here — it is `check_on_backend(backend, x, name)` in `common/gpu_utils.jl` next to the other backend primitives, with an `on_backend` predicate that `to_device` is the conversion for:

```julia
on_backend(::CPUBackend, x::AbstractArray) = x isa Array
on_backend(b::GPUBackend, x::AbstractArray) = x isa Base.typename(typeof(b.proto)).wrapper
```

src/compute_states.jl
jaemolihmbot: You are right that an error comes anyway: `el.e_full .= eigenpairs.e_full[:, ik]` broadcasts a length-`eig.nw` view into a length-`nw` vector, so a mismatch is a `DimensionMismatch` from inside the per-k copy. The guard only buys the message, and the fact that it fires once per call rather than inside a threaded loop.

I kept it on those grounds but moved the whole thing out of `compute_states.jl` in `710e707` — `_check_eigenpairs` now lives in `electron_eigenpairs.jl` next to the type it guards, and its residency half is the generic `check_on_backend` in `common/gpu_utils.jl`:

```julia
function _check_eigenpairs(eig::ElectronEigenpairs, nw, backend)
    eig.nw == nw || throw(ArgumentError(
        "eigenpairs holds nw = $(eig.nw) Wannier functions, but the model has nw = $nw"))
    check_on_backend(backend, eig.e_full, "eigenpairs")
    nothing
end
```

Say so if you would rather drop the `nw` line outright — it is one line and I have no attachment to it.

src/common/gpu_utils.jl
```diff
@@ -38,6 +38,31 @@ form. Not exported; use `ElectronPhonon.to_device`.
...
+
+"""
+    check_on_backend(backend, x::AbstractArray, name)
+
+Throw an `ArgumentError` naming `name` unless `x` is resident on `backend`. For a caller-supplied
+array that has to live on the same side as the run's own buffers: a mismatch is then reported here
+instead of surfacing deeper in as a mixed host/device operation, a per-element transfer, or -- on a
+path that happens to take `Array(x)` anyway -- no error at all.
+"""
+function check_on_backend(backend::AbstractBackend, x::AbstractArray, name)
```
jaemolihm: allow no name

src/compute_states.jl
jaemolihmbot: **Correction to my previous reply — the guard is not only about the message.** I said "an error comes anyway"; that is true for every mismatch but one, and I should have checked the exception before saying it.

`nbasis == 1` against a multi-band model **broadcasts**, silently:

```
dest length 5  .=  src length 1   ->  SILENTLY SUCCEEDED   (one value written into every band)
dest 5x5       .=  src 1x1        ->  SILENTLY SUCCEEDED
any other mismatch                ->  DimensionMismatch
```

So without the check, a 1-band cache handed to a multi-band run returns wrong numbers with no error. It needs two different models to trigger, which is why it never showed up, but it is a wrong-answer hole rather than a message-quality question. Keeping the guard on those grounds.

Current form after the rename (`36818fd`), unchanged in shape from what you saw:

```julia
function _check_eigenpairs(eig::Eigenpairs, nw, backend)
    eig.nbasis == nw || throw(ArgumentError(
        "eigenpairs holds nbasis = $(eig.nbasis), but the model has nw = $nw"))
    check_on_backend(backend, eig.e_full, "eigenpairs")
    nothing
end
```

src/compute_states.jl
jaemolihmbot: Still here after the rename, now as `check_on_backend(backend, eig.e_full, "eigenpairs")` inside `_check_eigenpairs` (`36818fd`). The type is `Eigenpairs` and the field is `nbasis` — see the rename commit; nothing about the residency argument changed.

src/common/gpu_utils.jl
jaemolihmbot: Done in `36818fd` — `name` defaults to `"array"`:

```julia
function check_on_backend(backend::AbstractBackend, x::AbstractArray, name = "array")
```

The eigenpairs call still passes `"eigenpairs"`, since there the specific word is what makes the message useful.

jaemolihm (approved):

PR #25, opened 2026-09-16
Split xk_to_ik into a checked entry point and xk_to_ik_unsafe

### LINKED ISSUE #26: state_index returns a wrong state index, not 0, for an xk off the selection's grid
`state_index(s, xk, iband)` documents *"or `0` if absent"* (`src/band_states.jl:441-452`), and the same docstring says the two selections' k-grids are **independent**, so only the k-vector is a shared address. But the lookup rounds `xk` onto `s.kpts`' grid, so an `xk` that is not a node of that grid resolves to a **neighbouring node** and the function returns a wrong nonzero state index instead of `0`.

## Reproduction

Two selections on grids that are not multiples of each other (4³ and 6³), querying every 6³ point that is *not* a node of the 4³ grid:

```julia
selA = FilteredBandStates(GridKpoints(kpoints_grid((4,4,4))), 1:64, fill(3, 64); nw = 8, nstates_base = 2.0)
for xk in GridKpoints(kpoints_grid((6,6,6))).vectors
    # skip the genuinely shared nodes
    state_index(selA, xk, 3)
end
```

| outcome | count |
|---|---|
| wrong nonzero state index (aliased) | **208** |
| `0`, as documented | 0 |

```
state_index(selA, [0.0, 0.0, 0.16666666666666666], 3) = 2, whose actual k is [0.0, 0.0, 0.25]
state_index(selA, [0.0, 0.0, 0.3333333333333333],  3) = 2, whose actual k is [0.0, 0.0, 0.25]
state_index(selA, [0.0, 0.0, 0.6666666666666666],  3) = 4, whose actual k is [0.0, 0.0, 0.75]
```

## Why it matters

`state_index` is exported, and the item form `state_index(s, other[i])` exists precisely to locate a state of *one* selection in *another* — the cross-grid case. `MigdalEliashberg.jl`'s `reindex_states` (`src/kernel_ci.jl`) is built on it, so the cross/interpolative kernel is a consumer.

## Candidate fixes (not picking one)

1. Validate the query is a node of `s.kpts`' grid and return `0` when it is not — i.e. the on-grid half of the checked `xk_to_ik` plus `nothing -> 0`. This is the *"off-grid possible + miss legitimate"* combination that neither `xk_to_ik` nor `xk_to_ik_unsafe` provides today, and several other call sites want it too (see #25's classification table).
2. Have the two grids agree by construction and assert it at the point where two selections are combined, leaving `state_index` a pure grid-node lookup.

## Scope

Deliberately **not** fixed in #25, which only renames the two lookups and classifies the call sites; `band_states.jl:461` stays on `xk_to_ik_unsafe` there so that PR is behaviour-preserving. Fixing this changes what `state_index` returns, which deserves its own change.

jaemolihmbot:
## Motivation

`xk_to_ik` hashes its argument onto the grid, so an **off-grid query returns a neighbouring node's index instead of missing**. That is a silent wrong answer, and it has already cost a debugging session: `superconductivity/dev/verify_fmp_doppler.jl:453-454` records it as a trap — an `xk_to_ik(k + Q/2, kpts)` test of *"does this shift map the grid onto itself"* always says yes, because the lookup rounds `k + Q/2` back onto the grid.

The rounding is still wanted in the pair loops, but it should not be what you get from the obvious name.

## Change

`src/common/kpoints.jl` — two exported names. The two ways a lookup can fail are not the same kind of event, so they no longer get the same answer:

| | off-grid query | node not in `kpts` | returns |
|---|---|---|---|
| `xk_to_ik` | **throws** — a malformed question | `nothing` — a legitimate answer | `Union{Int,Nothing}` |
| `xk_to_ik_unsafe` | rounds onto the nearest node | `nothing` | `Union{Int,Nothing}` |

**Routing rule: `xk_to_ik` everywhere, with two exceptions.**

| route | n | sites |
|---|---|---|
| `xk_to_ik` | **12** | everything not below |
| `xk_to_ik_unsafe` | **11** | the ten genuine (k, q) pair loops, where the validation is measurable, plus `check_model_symmetry.jl:74` |

`check_model_symmetry` is the one principled exception: it exists to diagnose a k-grid **not** closed under the point group, which is exactly the case that makes `S*xk` off-grid, so it must *report* that rather than throw. Its own comparison against the aliased node is an equivalent on-grid test with a message naming the grid — measured below.

Three sites treat a miss as a bug and now raise **their own** error on `nothing`, phrased in terms of what each was doing: the multigrid merge (`filter.jl:294`), `gamma_adaptive`'s per-state k index (`:114`), and the symmetry unfolding (`unfold.jl:102`, previously a `MethodError` on `isk_to_ik_isym[nothing]`).

`state_index` keeps its documented `0`-if-absent contract — absent meaning a k point the selection does not carry. An `xk` off its grid now throws instead of resolving to a neighbour.

Full per-site table with the call-count order for each: **see the comment below**.

## Benchmark

`@benchmark`, as a loop over a 110592-point (48³) k-list — the shape the hot loops have — on `workergpu068`, Julia 1.11.7:

| | min | median |
|---|---|---|
| `xk_to_ik_unsafe` | **21.1 ns/call** | 21.8 |
| `xk_to_ik` | **38.8 ns/call** | 39.4 |

**×1.84.** A single lookup in isolation gives 21.4 / 36.1 ns, i.e. the same picture. 17.7 ns/call is nothing at setup scale but ~16 s at the ~1e9 calls per run that `_wrap_reduced`'s own comment says the pair loops make — which is the whole reason the unchecked name survives. The figures are in `xk_to_ik_unsafe`'s docstring too.

## Test

New testset `"kpoints: xk_to_ik checked vs unsafe"` in `test/test_kpoints.jl`, **18/18**, pinning the difference rather than the happy path:

* own points and a `+1e-12` grid-arithmetic query — both names agree
* an off-grid query — checked throws; unsafe **aliases to `ik = 1`, asserted explicitly**
* half a grid spacing off — also rejected
* a node not in the set — **both** return `nothing`; the checked form does not throw there
* a shifted grid rejecting unshifted Γ while unsafe aliases it
* a lattice-vector-wrapped query resolving to the same point

Three pre-existing assertions that expected `nothing`, and `combine_kpoint_grids`' `@test ikq !== nothing`, use the unsafe name — against the checked form that last one would be a tautology.

Sweep on this branch (`test_kpoints` **16620/16620**): `test_iq_build` 4816, `test_filter_electron_states` 15, `test_band_states` 73, `test_screening` 10, `test_epmat` 9, `test_cubicBN_selfen` 24, `boltzmann/test_multigrid_weights` 193, `boltzmann/test_multigrid_transport` 91+3+3, `boltzmann/test_hdf5` 76+20, `test_ElectronState` 23, `test_diagonalize` 10, `test_velocity` 65, and `test_gpu.jl` every testset (CPU-vs-GPU rdiff 3.04e-16 / 8.11e-16) on an A100.

Known gap, pre-existing: `test_symmetry_operator.jl` and `boltzmann/test_el_transport_semiconductor.jl` cannot run — they need `test/data_cubicBN/epw_data_julia.bin`, which is in no checkout, and both are commented out in `runtests.jl`. Their sites were reviewed by reading.

Fixes #26 — `state_index` now throws on an off-grid `xk` instead of returning a wrong index, while still returning `0` for a k point the selection does not carry.


src/common/kpoints.jl
```diff
@@ -657,8 +658,41 @@ _hash_xk(xk, kpts::GridKpoints) = _hash_xk(xk, kpts.ngrid, kpts.shift)
...
+
+Use `xk_to_ik_unsafe` instead where `nothing` is a legitimate answer -- asking whether a point is in
+a filtered selection -- or inside a per-pair loop whose argument is a grid node by construction.
+"""
+function xk_to_ik(xk, kpts::GridKpoints{T}) where {T}
+    nxk = (xk - kpts.shift) .* kpts.ngrid
+    isapprox(round.(Int, nxk), nxk; atol = sqrt(eps(T))) || throw(ArgumentError(
+        "k point $xk is not on the grid of size $(kpts.ngrid) shifted by $(kpts.shift)"))
+    ik = xk_to_ik_unsafe(xk, kpts)
+    ik === nothing && throw(ArgumentError(
```
jaemolihm: on a node not in kpts -> nothing or throw ? isn't it better to let the caller deal with nothing ?

src/boltzmann/covariant_derivative.jl
```diff
@@ -199,7 +199,10 @@ function compute_covariant_derivative_matrix(el_irr::QMEStates{FT}, el_irr_state
 
         for (b, b_cart, wb) in zip(bvec_data...)
             xkb = xk + b
-            ikb = xk_to_ik(xkb, kpts)
+            # `_unsafe` is safe from aliasing here: `finite_difference_vectors` builds every `b`
```
jaemolihm: comment not needed

src/boltzmann/electron_master_equation.jl
```diff
@@ -460,7 +460,9 @@ function _qme_linear_response_unfold_map_nosym(el_i::QMEStates{FT}, el_f::QMESta
     for ind_el_i in 1:el_i.n
         ik = el_i.ik[ind_el_i]
         xk = el_i.kpts.vectors[ik]
-        ik_f = xk_to_ik(xk, el_f.kpts)
+        # On-grid by construction only as far as the assertions above reach: the `ngrid` one
```
jaemolihm: comment not needed

src/boltzmann/run_coherence.jl
```diff
@@ -294,7 +294,9 @@ function compute_electron_phonon_bte_data_coherence(model, btedata_prefix, windo
     else
         @views for ik = 1:nk
             xk = kpts.vectors[ik]
-            ik_kq = xk_to_ik(xk, kqpts)
+            # `kqpts` is built on a grid containing `kpts`, so `xk` is a node of it and cannot
```
jaemolihm: comment not needed

src/common/kpoints.jl
```diff
@@ -657,8 +658,41 @@ _hash_xk(xk, kpts::GridKpoints) = _hash_xk(xk, kpts.ngrid, kpts.shift)
...
+
+"""
+    xk_to_ik_unsafe(xk, kpts) -> Union{Int, Nothing}
+
+Index of the k point `xk` in `kpts`, `nothing` if `kpts` has none. Unchecked: `_hash_xk` rounds `xk`
+onto `kpts`' grid, so an off-grid argument returns the nearest node's index rather than `nothing`.
+Callers must therefore have established that `xk` is a node of that grid; [`xk_to_ik`](@ref) is the
+checked entry point that establishes it for them.
+"""
+function xk_to_ik_unsafe(xk, kpts)
```
jaemolihm, line 694: you should mention in docstring that one should use `xk_to_ik_unsafe` in hot loop for performance (and make sure it is safe usage). Also do `@benchmark` and compare timing and report it in PR and docstring.

src/postprocess/check_model_symmetry.jl
```diff
@@ -71,7 +71,11 @@ function check_electron_symmetry_of_model(model::Model{FT}, ngrid; fourier_mode=
         for ik in 1:kpts.n
             xk = kpts.vectors[ik]
             sk = symop.is_tr ? -symop.S * xk : symop.S * xk
-            isk = xk_to_ik(sk, kpts)
+            # `_unsafe` on purpose: on a grid not closed under `symop`, `sk` is off-grid and this
```
jaemolihm: no comment needed

src/band_states.jl
```diff
@@ -458,7 +458,9 @@ independent, so only the k-vector is a shared address.
 end
 function state_index(s::AbstractBandStates{T, <:GridKpoints},
         xk::Vec3, iband::Int) where {T}
-    ik = xk_to_ik(xk, s.kpts)
+    # `_unsafe` keeps today's behaviour, which is wrong for an `xk` off this selection's grid:
```
jaemolihm: in general : no comment needed for using xk_to_ik_unsafe unless it is unclear from context

src/common/kpoints.jl
jaemolihmbot: Done in 9edcd5e — `nothing`, and the caller deals with it.

`xk_to_ik` now throws only for an `xk` that is not a node of the grid (a malformed question) and returns `nothing` for a node the set does not hold (a legitimate answer). Return type `Union{Int,Nothing}`.

That collapsed the routing rule, which was the real gain: **checked by default**, unchecked only in a genuine per-pair loop and in `check_model_symmetry`, which must tolerate an off-grid query rather than throw on it. 12 sites checked, 11 unchecked (ten (k, q) pair loops plus that one) — previously 3 and 20.

Three sites treat a miss as a bug and now raise their own error, phrased in terms of what each was doing:

* `filter.jl:294` — names the merged grid and what it is supposed to hold
* `gamma_adaptive.jl:114` — names the state and the k point it sits at
* `unfold.jl:102` — names the symmetry image, the k point and the operation; this replaces a `MethodError` on `isk_to_ik_isym[nothing]`

Type stability re-checked: `xk_to_ik` infers `Union{Nothing,Int64}`, and inference narrows past the branch at both call shapes — `=== nothing && continue` and `=== nothing && error(...)` each leave `ik::Int64` with 0 `Any` slots and 0 `Core.Box`. `state_index` still infers a concrete `Int64`.

Two things this reroute catches that the old code did not:

* `state_index(s, xk, iband)` promised `0` if absent but returned a **wrong nonzero index** for an off-grid `xk` — 208 of 208 non-shared points when a 6³ selection is queried into a 4³ one. Now off-grid throws, absent still gives `0`. That closes #26.
* `electron_master_equation.jl:463` — the shift assertion above it is vacuous (#27), so a mismatched-shift `el_f` could reach the lookup off-grid; it now throws instead of aliasing.

Also worth flagging: `filter.jl:299` (the `ik_f` lookup) ended up **checked**, not unchecked — it is neither per-pair nor off-grid-tolerant, so the default applies. `test_kpoints` stays at 16620/16620 and the widened sweep is green.

src/boltzmann/covariant_derivative.jl
jaemolihmbot: Removed in 9edcd5e.

src/boltzmann/electron_master_equation.jl
jaemolihmbot: Removed in 9edcd5e.

src/boltzmann/run_coherence.jl
jaemolihmbot: Removed in 9edcd5e.

src/common/kpoints.jl
jaemolihmbot: Both done in 9edcd5e.

`xk_to_ik_unsafe`'s docstring now says it is the one to use in a hot loop once the caller has established that `xk` is a grid node, and carries the timing; `xk_to_ik`'s says a caller for which a miss is a bug raises its own error on `nothing`.

`@benchmark`ed as a loop over a 110592-point (48³) k-list — the shape the hot loops actually have — on `workergpu068`, Julia 1.11.7, 1 thread:

| | min | median |
|---|---|---|
| `xk_to_ik_unsafe` | **21.1 ns/call** | 21.8 |
| `xk_to_ik` | **38.8 ns/call** | 39.4 |

**×1.84.** A single lookup in isolation gives 21.4 / 36.1 ns, i.e. the same picture, so the loop figure is the one quoted. 17.7 ns/call is nothing at setup scale but ~16 s at the ~1e9 calls per run that `_wrap_reduced`'s own comment says the pair loops make — which is why the unchecked name survives at all. The numbers are in the docstring and in the PR body.

(This replaces the hand-rolled min-of-2 from the first round; the two agree to about 3%.)

src/postprocess/check_model_symmetry.jl
jaemolihmbot: Removed in 9edcd5e.

This site also keeps `xk_to_ik_unsafe`, and it is now the only non-per-pair site that does: the routine exists to detect a k-grid not closed under the point group, which is exactly the case that makes `S*xk` off-grid, so the checked form would throw where this routine wants to report. Its own guard at `:75-77` is already an equivalent on-grid test — over cubicBN's 48 operations on (3,4,4) and (2,3,4) it fired on 1824 and 920 off-grid images with zero false "symmetric" verdicts.

src/band_states.jl
jaemolihmbot: Removed in 9edcd5e, and taken as the general rule — every clause comment about `xk_to_ik_unsafe` is gone, including the four in `0cd3c70` and the three I added in `e0061fa`. **Nothing was kept anywhere.**

`filter.jl:276`/`:299` was the one pair I would have argued for, since its argument rests on the `all(mod.(nks_f, nks_c) .== 0)` guard 16 lines up rather than on anything at the site. It turned out moot: under the new contract both are neither per-pair nor off-grid-tolerant, so they take the checked `xk_to_ik` and there is no unchecked usage left there to explain. So the question does not arise — but if you would still rather those two were unchecked with a comment, say so and I will swap them back.

This site itself is now checked: `state_index` keeps its documented `0`-if-absent contract, where absent means a k point the selection does not carry, and an `xk` off its grid throws rather than resolving to a neighbour. That is the fix for #26, which the docstring now states.

jaemolihm (approved):

PR #28, opened 2026-09-16
Let the e-ph drivers take their electron eigenpairs from a shared cache

jaemolihmbot:
Stacked on #23 (`feat/electron-eigenpairs-consume`). Merge order: #25 → #22 → #23 → this.

The design document this implements is attached as a [provenance comment](https://github.com/jaemolihm/ElectronPhonon.jl/pull/28#issuecomment-5700195375) (`plans/` is local and never committed).

## Motivation

`compute_electron_states` can already take an `ElectronEigenpairs` cache (#23), but nothing above it
can pass one. The e-ph drivers are where it matters: two separate runs over *overlapping* k-sets
each diagonalize H(k) themselves, so a band-resolved `g2 = |g|²` — which is basis-dependent inside a
degenerate multiplet — is only comparable between them by coincidence. The consumer is the
cross/interpolative kernel, which multiplies a block cut from one restricted run by the
pseudo-inverse of a block cut from a *different* one.

This makes that agreement structural: both runs are handed the same cache and copy their eigenpairs
out of it. The per-k CPU eigensolve pins a multiplet basis only through the EPW-mimicking fix in
`solve_eigen_el!`, and the batched device eigensolve does not pin it at all.

## Change

- `_setup_electron_k(...; el_k_eigenpairs = nothing)` — one edit serving **all three** drivers.
- `el_kq_eigenpairs` on `run_eph_over_k_and_kq` only, through `_setup_electron_kq` and
  `_compute_electron_states_kq` (both compute sites, so the unfolding path is served too).
- `run_eph_over_k_and_q` / `run_eph_over_q_and_k` get the k side only. Their k+q states stay
  unwired — D3 = (i). No positional `Union` changed: a cache is not a k-point specification.
- Docstring on `run_eph_over_k_and_kq`: why the keywords exist, per-rank coverage (the caches are
  **not** MPI-distributed; `mpi_comm_k` splits the outer k inside `filter_electron_states`), a miss
  is an error, read-only, backend-residency.
- D6 = (i): `el_kq_eigenpairs` **composes** with `el_kq_from_unfolding = true` — documented, not
  guarded. Runs sharing a k+q cache must make the *same* choice: at one full-BZ k+q point an
  unfolding run carries the cached *irreducible* eigenvector rotated by the symmetry operation while
  a direct run carries the cached one at that point, and nothing checks it.

## Test

New `test/test_eph_precomputed_states.jl`. Pb 4³, outer selections `1:48` and `25:64` (24 shared),
`fourier_mode` passed explicitly to both arms of every comparison.

**The discriminating check is a deliberately wrong cache** — the same eigenpairs rotated by one
k point, so it is well-formed, covers every k point and is resident on the run's own backend, but
holds the wrong eigenvector at each:

```julia
perm = [mod1(ik + 1, kgrid.n) for ik in 1:kgrid.n]
bad  = ElectronEigenpairs(model.nw, kgrid, cache.e_full[:, perm], cache.u_full[:, :, perm])
```

| branch | max abs Δ`u_full`, CPU | GPU | max abs Δ`e_full` |
|---|---|---|---|
| outer k | 1.6189 | 1.6188 | 0.5313 Ha |
| inner k+q | 1.6189 | 1.6188 | 0.5313 Ha |
| prebuilt k+q | 1.6189 | 1.6188 | — |

**With the keyword deleted the same predicates read `0.0`.** That is what makes this file a test:
the provenance checks (`u_full == cache.u_full[:, :, ik]`), the one-k-offset control and the
shared-gauge assertion all pass vacuously without it, because on this fixture H(k) comes out bitwise
independent of which other k points a run visits. A `fourier_mode`-mismatched cache is a second
tooth on CPU (max abs Δ`u` = 1.5716) but is inert on GPU, where the cache builder and the device
eigensolve both hardcode `fourier_mode = "batched"`.

Also covered: inertness (a cached run is bitwise equal to the uncached one over `xk`, `e_full`,
`u_full`, `nband`, `rng`, `vdiag`, `v`, `rbar`, on both sides and both backends); the shared-gauge
deliverable `(agreement, n_shared) == (true, 24)` with its failing counterpart for the wrong cache; a
miss teeth per side, so each keyword's forwarding is covered alone; the prebuilt-`FilteredBandStates`
k+q branch, which the grid runs never enter and the cross driver always takes.

**Tally: 39/39 with CUDA, 23/23 without — the 16 GPU-gated assertions account for the difference
exactly.** Regression sweep green (`test_electron_eigenpairs` 85/85 & 57/57, `test_kpoints`
16620/16620, `test_gpu.jl` every testset, `test_cubicBN_{spectral,selfen}`, the BTE and multigrid
files).

## Two honest limits

- **On this Pb fixture the feature fixes nothing.** Two *uncached* runs over the differing k-sets
  already agree bitwise, on CPU and GPU alike (max abs Δ`u_full` = 0.0 at all 24 shared points; also
  0.0 for device `compute_electron_states` over 400-vs-300 k on 8³ and 1500-vs-1000 k on 12³). The
  plan predicted the GPU arm would fail today; that prediction was wrong, and the review confirmed
  it. This is a structural guarantee for the Kramers / degeneracy-pervasive class, where every band
  is doubly degenerate at every k and every `g2` the cross touches becomes basis-dependent.
- **The D6 unfolding composition is wired and documented but unexercised in-repo.**
  `unfold_ElectronStates` reads `model.el_sym.operators`, and `el_sym === nothing` for both artifact
  models, so that path throws with or without a cache — `test/test_unfold.jl` is commented out of
  `runtests.jl` for the same reason. Pinned by an assertion in the new test file rather than left as
  prose.


src/calculator/run_eph_over_k_and_kq.jl
```diff
@@ -16,6 +16,33 @@ coupling. Two keywords select where the arrays live and which payload the calcul
 The batched path has a narrower scope than the per-point one (no polar/long-range, no screening,
 `energy_conservation = (:None, 0.0)`, commensurate grids, no `covariant_derivative_of_g`, no
 `skip_eph`, no `el_kq_from_unfolding` under symmetry); it asserts each of these.
+
+Two further keywords let a run take its electron eigenpairs from a shared cache instead of
```
jaemolihm, line 20: comment toooo long

jaemolihmbot: Cut from 26 lines to 13 in `2aa3916`. What stayed: the purpose (shared gauge, why `g2` needs it), per-rank coverage, backend residency, read-only, and the `el_kq_from_unfolding` gauge trap. What went: the MPI elaboration, the restatement of what a consumer copies out, and the longer version of the unfolding argument.

Also gave `_setup_electron_k`/`_setup_electron_kq` the `Union{Nothing, ElectronEigenpairs}` kwarg annotation the public drivers already had.

The branch is restacked on the updated #23 (which grew `710e707`); `test_eph_precomputed_states.jl` 39/39 and `test_electron_eigenpairs.jl` 85/85 with `CUDA.functional() == true`, so the GPU arms ran.

src/calculator/eph_setup_common.jl
```diff
@@ -9,12 +9,15 @@
 # passed through verbatim, or filtered from a grid to the energy window) and compute the electron
 # states there. `backend` says where the eigensolves run (a non-CPU backend takes the batched device
 # path). `verbosity` selects timing (@time when > 0, via `maybe_time`); `el_k_quantities` lets
-# the outer-q batched path skip velocity/position. The prebuilt pass-through is dead for the two
-# grid-only sibling callers (they always pass a grid), so their behavior is unchanged.
+# the outer-q batched path skip velocity/position. `el_k_eigenpairs` is an optional
```
jaemolihm: 개조식으로 해주세요

jaemolihmbot: `760d2c5`. 두 블록 모두 개조식으로 바꿨습니다 — `_setup_electron_k`와, 바로 아래 형제인 `_compute_electron_states_kq`도 같이 (한 파일 안에서 한쪽만 개조식이면 어색해서). 각각 리드 한 줄 + knob 당 불릿 하나입니다.

이 브랜치는 이름 변경된 #23 위로 restack 했습니다 (`ElectronEigenpairs` → `Eigenpairs`, `nw` → `nbasis`). 이름 변경은 스택 각 커밋에 접어 넣어서, 중간 커밋도 전부 컴파일됩니다. `CUDA.functional() == true` 상태에서 `test_eigenpairs.jl` 85/85, `test_eph_precomputed_states.jl` 39/39.

PR #30, opened 2026-09-17
Let filter_electron_states take the window test's eigenvalues from an Eigenpairs cache

jaemolihmbot:
Stage 4 of the eigenpair-cache series (#22 → #23 → #28), on top of the merged stack.

## Motivation
`Eigenpairs` exists so that several e-ph runs over overlapping k-point sets share one eigenvector
gauge. `filter_electron_states`' window test was still diagonalizing `H(k)` itself — the last
independent eigensolve of the same Hamiltonian in a run. Passing it the cache removes that solve,
but the real gain is **consistency**: the same eigenvalues then decide the window *and* end up in
the states of a `compute_electron_states(…; eigenpairs)` run over the same cache, rather than coming
from two different eigensolves that agree only to round-off.

## Change
`src/filter.jl`
- `filter_electron_states(…; eigenpairs::Union{Nothing, Eigenpairs} = nothing)`, with the existing
  `_check_eigenpairs` at the entry (`nbasis` vs the model's `nw`, plus backend residency), forwarded
  through `_filter_with_band_ranges` on **both** the serial and the `mpi_comm` branch.
- `_filter_kpoints` reads the cache on both of its paths, as `::Nothing` / `::Eigenpairs` method
  pairs — so neither loop gains a branch, and with a cache **no interpolator is built at all**:
  - `_set_eigenvalues_from!` — threaded per-k; cached column copied into the thread-local buffer
    instead of `get_el_eigen_valueonly!`
  - `_eigenvalues_on_host` — device path; a gather off the resident cache instead of
    `get_el_eigen_valueonly_batched`
- Semantics documented in the docstring: the cache must cover the **input** grid, not the filtered
  result (the window test is what needs the energies); it must be resident on `backend`; it is not
  MPI-distributed, so under `mpi_comm` build it over the whole grid on every rank; a k-point it does
  not hold throws rather than silently recomputing.

`src/ElectronPhonon.jl` — `include("eigenpairs.jl")` moved above `include("filter.jl")` (it needs
only `GridKpoints` and `Model` at load time).

## Test
Pb artifact, 12³ = 1728 k, `E_F ± 0.2 eV` → 132 k, `band_range = 2:3`, per-k extent varying over
`2:2` / `3:3`. New `@testset "consumed by filter_electron_states"` in `test/test_eigenpairs.jl`,
next to the Stage-2 `compute_electron_states` one so the cache's consumers stay together.

- **Window decision bitwise identical with and without a cache, on both backends** — field-by-field
  `==` on `kpts.vectors`, `iks`, `ibands`, `band_extent`, `nstates_base`. All four arms
  (CPU/GPU × cache/no-cache) give the identical selection.
- **The payoff, asserted:** for every retained k, `band_extent[ik] == inside_window(cache.e_full[:, j], window...)`
  *and* `compute_electron_states(model, sel, …; eigenpairs = cache)[ik].e_full == cache.e_full[:, j]`
  at `j = xk_to_ik(xk, cache.kpts)`. Decision and state energies are the same numbers by construction.
- **Teeth:** a cache with its k columns rotated by one is well-formed and covers the grid, and moves
  the selection. (A Stage-3-style `fourier_mode` mismatch would *not* work here — it moves `e` by
  1e-15, which no window decision can see.)
- **Coverage / residency / MPI:** a cache over the *filtered* 132-point set — the natural mistake —
  throws `"eigenpairs does not cover k point …"` on both backends; `nbasis` and residency throw at
  the entry, which matters most on this path because the eigenvalue-only device path would otherwise
  consume a host cache with no error at all. `mpi_comm = MPI.COMM_SELF` with a whole-grid cache
  reproduces the serial cached result.
- **Type stability:** `_filter_kpoints` infers the same fully concrete `NamedTuple` and has 0
  `Core.Box` with and without a cache.

Results, `CUDA.functional() = true` so both GPU arms ran:

| file | this branch | at base `23aa72e` |
|---|---|---|
| `test/test_eigenpairs.jl` | **99/99** | 85/85 |
| `test/test_filter_electron_states.jl` | **15/15** | 15/15 |
| `test/test_kpoints.jl` (shared `filter.jl`) | 16/16/9/147/18/16403/11 | — |
| `test/boltzmann/test_multigrid_weights.jl` | 193/193 | — |

## Caveat, stated rather than papered over
The window *decision* is bitwise identical, but the energies behind it are not: the filter's
value-only solve and the cache's full eigensolve differ by `max|Δe| = 8.88e-16` aru (1.21e-14 eV)
over all 1728×4 eigenvalues. So a band sitting within ~1e-15 of a window edge could legitimately
fall on the other side of it with a cache than without. None does on this fixture. This is in the
docstring and the test comment.

## Not done
Not extended to `filter_electron_states_multigrid` (one fine-grid cache could serve both of its
grids, since coarse nodes are fine nodes) or to `filter_qpoints` — the plan's Stage 4 names
`filter_electron_states` only.


src/filter.jl
```diff
@@ -37,7 +37,34 @@ end
 filter_kpoints(kpts_input, nw, el_ham, window, mpi_comm; kwargs...) =
     filter_kpoints(kpts_input, nw, el_ham, window; mpi_comm, kwargs...)
 
-function _filter_kpoints(nw, kpoints, el_ham, window; fourier_mode="normal", backend=CPUBackend())
+# The full-band eigenvalues of one k point, written into `eigenvalues`: solved on the spot, or
+# copied out of a supplied cache. The cache arrives as a typed argument, so each call site
+# specializes on one of the two methods and the no-cache path is the plain value-only solve it was.
+_set_eigenvalues_from!(eigenvalues, nw, ::Nothing, ham, xk) =
```
jaemolihm: I thought  this was defined somewhere else as well ? I think move this to eigenpairs.jl

jaemolihmbot: Yes — `compute_states.jl` had the same solve-or-copy-out pair for the full eigenpair, and the `"eigenpairs does not cover k point"` lookup was written out at **five** sites. Moved in `60d6f6a`.

`eigenpairs.jl` now owns all four method pairs over one lookup helper:

```julia
function _eigenpairs_ik(eigenpairs::Eigenpairs, xk)   # the lookup + the miss error, once
    ik = xk_to_ik(xk, eigenpairs.kpts)
    ik === nothing && throw(ArgumentError("eigenpairs does not cover k point $xk"))
    ik
end

_set_eigen_from!           # full eigenpair into an ElectronState   (was compute_states.jl)
_set_eigen_valueonly_from! # eigenvalues into an ElectronState      (was compute_states.jl)
_set_eigenvalues_from!     # eigenvalues into a plain buffer        (was filter.jl)
_eigenvalues_on_host       # batched gather for the window test     (was filter.jl)
```

`grep -c "does not cover k point" src/` is now 1 (was 5) — the device gather in `_compute_electron_states_device!` routes through it too.

One consequence worth flagging: `include("electron_state.jl")` had to move **above** `include("eigenpairs.jl")`, because the consumer methods dispatch on `ElectronState` and `eigenpairs.jl` must still precede `filter.jl`, which annotates an `Eigenpairs` kwarg. So the include order is now `model.jl → electron_state.jl → eigenpairs.jl → filter.jl`, with a comment saying why.

Verified in a fresh session (the first run silently reused a stale loaded module, so I restarted and confirmed all four helpers report `eigenpairs.jl` via `methods(...)` before trusting anything). `CUDA.functional() = true`: `test_eigenpairs.jl` 99/99, `test_filter_electron_states.jl` 15/15, `test_eph_precomputed_states.jl` 39/39, `test_ElectronState.jl` 23/23.

src/ElectronPhonon.jl
```diff
@@ -41,8 +41,11 @@ include("longrange.jl")
 include("wannier_to_bloch.jl")
 include("wannier_to_bloch_batched.jl")
 include("model.jl")
-include("filter.jl")
 include("electron_state.jl")
+# Before filter.jl, which annotates an `Eigenpairs` kwarg, and after electron_state.jl, whose
```
jaemolihm: drop comment

jaemolihmbot: Dropped in `a8da371`. The include list carries no other comments, so it was noise.

PR #31, opened 2026-09-17
Add holstein_model, an analytic Model builder

jaemolihmbot:
Build a `Model` for the Holstein model directly, without an EPW run: a single-orbital nearest-neighbor band on a 1d/2d/3d (hyper)cubic lattice coupled to a dispersionless Einstein phonon with a constant `g`.

```
Ĥ = ε₀ ∑ ĉ†ĉ - t ∑_⟨R,R'⟩ ĉ†_R ĉ_R' + ω₀ ∑ b̂†b̂ + g ∑ ĉ†_R ĉ_R (b̂_R + b̂†_R)
⟹ ε_k = ε₀ - 2t ∑_{i≤dim} cos(k·a_i),   ω_q = ω₀,   g_{mn,ν}(k,q) = g
```

`holstein_model` is `public` but not exported. It fills every field `load_model_from_epw_new` does, so the `Model` drops into the existing drivers, symmetry reduction included.

## Two representation choices worth calling out

**`nmodes = 1`, a scalar Einstein coordinate** rather than three Cartesian displacements of one atom. A coupling to a Cartesian displacement of an atom on a high-symmetry site is odd under the point group, so it could never be both constant and compatible with the symmetry group spglib reports — a totally symmetric scalar mode can. Consequently `model.mass` has length `nmodes = 1` while `structure.mass` keeps the crystal 3-per-atom convention; that asymmetry is commented at the `Model(...)` call.

**For `dim < 3` the unconnected directions get `100·alat`.** Besides isolating the wire/sheet, this lowers the symmetry group to one the `dim`-dimensional band actually obeys. A cubic cell would make spglib report operations mixing `x` and `z`, and symmetry-reduced calculations would then be silently wrong.

## Other changes

- `Model` gains a `dim` field, defaulted to 3 so both EPW loaders are unaffected. Nothing consumes it yet: transport and occupation routines still take `volume` explicitly from the caller, so a 2d conductivity still carries the arbitrary `100·alat` thickness unless the caller normalizes.
- Julia compat moves to 1.11 for the `public` keyword.

## Verification

`test/test_holstein_model.jl` (950 assertions, ~2 s, no artifacts needed). Against closed forms for `dim ∈ {1,2,3}` × `epmat_outer_momentum ∈ {el, ph}`:

| check | agreement |
|---|---|
| `ε_k` | 8e-17 |
| velocity | 6e-16 |
| `ω_q = ω₀` | exact |
| `g2 = \|g\|²` | 1e-19 |

Plus: `:Direct` and `:BerryConnection` velocities agree bitwise (position matrix elements vanish); `ε(Sk) = ε(k)` for every one of the 32 (1d/2d) / 96 (3d) reported symmetry operations; and `g2` stays constant through `run_eph_over_k_and_kq` with symmetry reduction (36 → 10 IBZ k-points).

`test_velocity.jl` and `test_epmat.jl` still pass with the `Model` change.


src/holstein_model.jl
```diff
@@ -0,0 +1,179 @@
+# Holstein model: a single-orbital tight-binding band coupled to a dispersionless
+# (Einstein) phonon with a momentum-independent coupling constant.
+
+public holstein_model
```
jaemolihm: put this in models/ (later we could add more models)

src/holstein_model.jl
```diff
@@ -0,0 +1,179 @@
...
+
+# Atomic mass used for the Einstein oscillator, in Rydberg atomic units (1 amu = 911.444).
+# The value is arbitrary: `Model` stores the e-ph matrix in the Cartesian displacement basis,
+# where the mass enters as epmat ∝ √M and the phonon eigenvector as 1/√M, so M cancels
+# exactly from `ω_q` and from every e-ph matrix element. It is fixed here (rather than
+# exposed) so that no parameter of `holstein_model` is physically inert.
+const HOLSTEIN_MASS = 911.444
```
jaemolihm: maybe just mass = 1 and no const, or pass mass to holstein_model and default 1.0 if you want to check mass dependence

src/holstein_model.jl
```diff
@@ -0,0 +1,179 @@
...
+
+"""
+    holstein_model(; t, ω₀, g, alat = 1.0, ε₀ = 0.0, dim = 3,
```
jaemolihm: - dim -> dimension (also in Model)
- g = \sqrt(2 * dimension * lambda * w0 * t). Allow passing either g or lambda (but not both)

and when printing print both g and lambda

note lambda is dimensionless (unitless) parameter

src/holstein_model.jl
```diff
@@ -0,0 +1,179 @@
...
+    vacuum = FT(100)
+    lattice_diag = Vec3{FT}(ntuple(i -> i <= dim ? alat : vacuum * alat, 3))
+    lattice = Mat3{FT}(Diagonal(lattice_diag))
+
+    atom_pos = [zero(Vec3{FT})]  # alat units, Cartesian
+    structure = Structure(alat, lattice, [mass], atom_pos, ["A"])
+
+    # --- Real-space R vectors ----------------------------------------------------------
+    # Electrons: on-site plus the 2*dim nearest neighbors. Phonon and e-ph: on-site only.
+    irvec_el = [Vec3{Int}(ntuple(j -> j == i ? s : 0, 3)) for i in 1:dim for s in (-1, 1)]
```
jaemolihm: can you just be explicit ? and do if branch on dimension

src/holstein_model.jl
```diff
@@ -0,0 +1,179 @@
...
+
+    # A single Wannier function at the origin: ⟨0|r|R⟩ = 0 for every R.
+    el_pos = WannierObject(irvec_el, zeros(Complex{FT}, nw^2 * 3, length(irvec_el)))
+
+    # With vanishing position matrix elements the Berry-connection velocity reduces to
+    # dH/dk, which is exactly what `el_ham_R` interpolates. `el_vel` is a copy of it (not the
+    # same object, to avoid aliasing on `update_op_r!`) so that `el_velocity_mode = :Direct`
+    # is also available and gives identical results.
+    el_vel = WannierObject(irvec_el, copy(el_ham_R.op_r))
+    el_velocity_mode = :BerryConnection
```
jaemolihm: :Direct

src/holstein_model.jl
```diff
@@ -0,0 +1,179 @@
...
+    else
+        # op_r indexed as (iw, jw, imode, Rₑ, Rₚ)
+        data = reshape(epmat, nw^2 * nmodes * length(irvec_el), length(irvec_ep))
+        ep = WannierObject(irvec_ep, data; irvec_next = irvec_el)
+    end
+
+    Model(; structure.alat, structure.lattice, structure.recip_lattice, structure.volume,
+        dim, nw, nmodes,
+        wann_centers = [zero(Vec3{FT})],
+        # `mass` is per phonon mode (length nmodes), not per atom: the scalar Einstein
```
jaemolihm: no comment needed

src/model.jl
```diff
@@ -32,6 +32,12 @@ Base.@kwdef mutable struct Model{FT <: AbstractFloat, WannType <: Union{Nothing,
     recip_lattice::Mat3{FT}
     volume::FT
 
+    # Spatial dimension of the system (1, 2 or 3). The lattice is always 3d; `dim < 3` means
+    # the remaining directions are decoupled vacuum, so quantities normalized per `volume`
+    # (carrier density, conductivity) refer to the full 3d cell and must be rescaled to a
+    # length or area by the caller.
+    dim::Int = 3
```
jaemolihm: dim -> dimension

src/holstein_model.jl
jaemolihmbot: Done in b968e8f — `git mv`d to `src/models/holstein_model.jl`, include updated.

src/holstein_model.jl
jaemolihmbot: Done in b968e8f — took the second option: `mass` is now a kwarg defaulting to `1.0`, and the module constant is gone. Added a "mass cancels out" testset that builds the model at `mass = 1.0` and `911.444` and checks `ω_q` and `g2` are unchanged.

src/holstein_model.jl
jaemolihmbot: Done in b968e8f, all three parts.

- `dim` → `dimension`, in the kwarg and in the `Model` field.
- `g` and `λ` are now alternatives — exactly one, or it is an `ArgumentError` (both given and neither given both error). `g = √(2·dimension·λ·ω₀·t)`, so `λ = g²/(2·dimension·ω₀·t) = g²/(ω₀·W/2)` against the half-bandwidth `W/2 = 2·dimension·t`; documented as such and noted as dimensionless. Deriving `g` from `λ` needs `t > 0`, which errors rather than silently giving `g = 0`.
- Both are printed with the rest of the parameters. Used `println` behind a `verbose` kwarg defaulting to `true`, so the tests can silence it:

```
Holstein model: dimension = 3, alat = 1 bohr
  t = 0.1 Ry, ε₀ = 0 Ry, ω₀ = 0.01 Ry, mass = 1
  g = 0.0547723 Ry, λ = 0.5
```

A new testset checks the round trip: building from `λ` and from the corresponding `g` gives the same `epmat`.

src/holstein_model.jl
jaemolihmbot: Done in b968e8f — the R vectors are an explicit `if dimension == 1 / 2 / 3` listing them literally, in the `reverse(R)` order `WannierObject` requires (its constructor checks the sort, so a typo fails loudly). Made the lattice explicit the same way instead of the `ntuple`.

src/holstein_model.jl
jaemolihmbot: Done in b968e8f — `el_velocity_mode = :Direct` now. The test asserts the default and still checks `:Direct` and `:BerryConnection` agree bitwise (position matrix elements vanish).

src/holstein_model.jl
jaemolihmbot: Done in b968e8f — comment removed. The point is still in the docstring under "Representation inside the `Model`".

src/model.jl
jaemolihmbot: Done in b968e8f — `Model.dimension`.

src/models/holstein_model.jl
```diff
@@ -0,0 +1,215 @@
...
+        "epmat_outer_momentum must be \"el\" or \"ph\", got $epmat_outer_momentum"))
+    (g === nothing) == (λ === nothing) && throw(ArgumentError(
+        "pass exactly one of `g` and `λ`, got g = $g and λ = $λ"))
+
+    FT = Float64
+    t, ω₀, mass, alat, ε₀ = FT(t), FT(ω₀), FT(mass), FT(alat), FT(ε₀)
+
+    # λ is measured against the half-bandwidth 2·dimension·t.
+    if g === nothing
+        t > 0 || throw(ArgumentError("deriving g from λ needs t > 0, got t = $t"))
```
jaemolihm: use abs(t)

src/models/holstein_model.jl
```diff
@@ -0,0 +1,215 @@
...
+        println(@sprintf("Holstein model: dimension = %d, alat = %g bohr", dimension, alat))
+        println(@sprintf("  t = %g Ry, ε₀ = %g Ry, ω₀ = %g Ry, mass = %g", t, ε₀, ω₀, mass))
+        println(@sprintf("  g = %g Ry, λ = %g", g, λ))
+    end
+
+    nw = 1
+    nmodes = 1
+
+    # --- Crystal structure -------------------------------------------------------------
+    # Directions without hopping are stretched so that the cell is an isolated wire
```
jaemolihm, line 141: can we filter symmetry after spglib call to keep only 1d or 2d ones ? Or does spglib (recent version) support low dim systems ?

test/test_holstein_model.jl
```diff
@@ -0,0 +1,175 @@
+using Test
+using LinearAlgebra
+using Random
+using ElectronPhonon
+using ElectronPhonon: holstein_model, AbstractCalculator, OuterKLoop, EPData, OuterIteration
+
+# A minimal calculator recording the range of g2 and ω seen by the driver. The driver calls
+# `run_calculator!` from several threads, so the accumulation is guarded by a lock.
+mutable struct _HolsteinG2Calc <: AbstractCalculator
```
jaemolihm: why this new struct ??? seems weird to have

src/models/holstein_model.jl
jaemolihmbot: Done in e83d88f — the half-bandwidth is now `2·dimension·abs(t)`, used in both directions:

```julia
half_bandwidth = 2 * dimension * abs(t)
g = sqrt(λ * ω₀ * half_bandwidth)        # from λ
λ = g^2 / (ω₀ * half_bandwidth)          # from g
```

The guard became `t ≠ 0` instead of `t > 0`, and the docstring says `W = 4·dimension·|t|`. A test checks that `t` and `-t` give the same `epmat` for a given `λ`.

src/models/holstein_model.jl
jaemolihmbot: Kept the vacuum cell, per your call.

For the record, on the two halves of the question:

- **spglib low-dim**: no. The pinned Spglib.jl exposes no layer-group API (no `get_layer_dataset`); `symmetry_operations` only has the 3d path.
- **Filtering after the call**: it does work. Keeping the cubic-cell operations whose `S` does not mix the hopping directions with the inactive ones gives, for dim = 1/2/3, 96 → 32/32/96 — the *identical* `(S, is_tr)` set the elongated cell produces, and closed under composition. It would have bought `volume = alat³` instead of `100^(3-dimension)·alat³`. The cost was that `model.symmetry` would then disagree with `model.structure.symmetry` (which stays the full cubic group, since `Structure` computes its own), so the vacuum cell is the self-consistent one.

test/test_holstein_model.jl
jaemolihmbot: Fair — replaced in e83d88f with a real `BoltzmannCalculator` run, and the struct is gone.

The reason a calculator was there at all is that `run_eph_over_k_and_kq` returns only `(; kpts, qpts, el_k_save, el_kq_save, ph_save)`; `g2` lives in the inner loop and `run_calculator!` is the only way out. But the phonon check needed no calculator (it reads `res.ph_save`), and `g2` was already verified directly for every dim × `epmat_outer_momentum`, so the struct was earning very little.

`BoltzmannCalculator` gives something much better. With `|g|²` and `ω` constant the scattering-out rate collapses onto the DOS,

Γ_k = 2π|g|² ∑_± (occupation) · D(ε_k ± ω₀)

so two properties hold exactly, with no dependence on the prefactor/occupation convention:

1. **Γ depends on k only through ε_k.** Not implied by symmetry — on the 6×6 grid the IBZ has an accidental degeneracy between the symmetry-inequivalent (0, ½, 0) and (⅙, ⅓, 0), both at ε = 0. Their Γ agree to **0.0** relative difference while differing from their neighbours by two orders of magnitude (0.0022 vs 0.246), so it is a sharp check rather than a coincidence.
2. **Γ ∝ |g|².** Doubling `g` gives exactly 4× Γ, again to 0.0 relative difference.

Plus a guard that Γ is not trivially constant across states, so neither check can go vacuous.

src/models/holstein_model.jl
jaemolihmbot: Done in a7bd7da — vacuum cell kept, and the symmetry is now lowered explicitly on top of it.

`_restrict_symmetry_to_dimension` drops any operation whose rotation mixes the `dimension` hopping directions with the inactive ones (the lattice is diagonal, so this reads straight off `S`). `model.symmetry` is that subgroup; `model.structure.symmetry` stays spglib's own result.

Worth knowing: **on the elongated cell this removes nothing.** spglib already returns exactly this subgroup — 32 / 32 / 96 operations for dimension 1 / 2 / 3, removed = 0 in every case. So it is a guarantee, not a behavioral change: `model.symmetry` can no longer exceed the symmetry of the `dimension`-dimensional Hamiltonian whatever spglib makes of a 100:1 cell. The case where it would bite is the vacuum factor being degenerate with `alat` — a cubic cell, where it cuts 96 → 32.

Tested both directions, so nothing here is vacuous:
- `model.symmetry` is block-diagonal, a subgroup of `structure.symmetry`, and closed under composition;
- the helper is exercised directly on a cubic cell, where it does cut 96 → 32 / 32 / 96.

src/models/holstein_model.jl
```diff
@@ -0,0 +1,242 @@
...
+inactive ones, i.e. whose rotation is block diagonal between `1:dimension` and the rest.
+
+The Hamiltonian built here has no hopping along the inactive directions, so any operation
+mixing the two blocks would not be a symmetry of it. The elongated cell already makes spglib
+report exactly this subgroup, so in practice nothing is removed; doing it explicitly means
+`model.symmetry` can never exceed the symmetry of the `dimension`-dimensional Hamiltonian,
+whatever spglib makes of a 100:1 cell. The lattice is diagonal, so the crystal and Cartesian
+axes coincide and the condition can be read off `S` directly.
+"""
+function _restrict_symmetry_to_dimension(symmetry::Symmetry, dimension::Integer)
```
jaemolihm: this should live in symmetry, and should be called without prepending _

one could even do the symmetry generation function with optional dimension

and it should check lattice is correct, i.e. for dimension - non-dimension off-diagonal should be 0

jaemolihmbot: Done in 89ad42e, all three parts.

- Moved to `src/common/symmetry.jl` as `restrict_symmetry_to_dimension(symmetry, dimension, lattice)`, exported, no leading underscore.
- **Lattice check added.** This was a real hole, not just hygiene: the block-diagonal test reads off `S` in reciprocal *crystal* coordinates, so it only constrains the Cartesian directions when the lattice itself does not mix the blocks. `lattice[i, j]` must now vanish whenever exactly one of `i, j` is `<= dimension`, else `ArgumentError`. A skewed lattice used to be filtered silently on the wrong basis; there is a test for it now.
- **Optional `dimension` on the generation function**, as you suggested: `symmetry_operations(lattice, atoms; dimension = 3)` applies the restriction, and `Structure` forwards its own `dimension` keyword to it. Default 3 everywhere, so no existing caller changes.

Nice side effect: `holstein_model` now just passes `dimension` to `Structure` and hands `structure.symmetry` to the `Model` as before, so `model.symmetry === model.structure.symmetry` again — the disagreement I flagged last round is gone.

Tests: the restriction is exercised both through `restrict_symmetry_to_dimension` directly and through the `Structure` keyword (cubic cell, 96 → 32/32/96), plus the skewed-lattice and bad-dimension errors. `test_symmetry.jl` still passes, since this touches core symmetry code.

src/models/holstein_model.jl
```diff
@@ -1,9 +1,21 @@
...
+# reused here: it lives in a different package (EPSpectral depends on nothing from here), works in
+# Hartree units, is 1d-only, and carries its own `μ` and `T`. Most of all it is not an
+# ElectronPhonon `Model` — it feeds EPSpectral's own spectral-function solver directly, whereas the
+# point of this builder is to produce a `Model` that the Wannier interpolation and e-ph drivers
+# consume like any EPW-loaded one.
 
 using Printf
 
-public holstein_model
+# `public` (Julia >= 1.11) marks the name as supported without exporting it. Gated exactly as in
```
jaemolihm: add julia = "1.11" to both packages and remove gate

test/test_holstein_model.jl
```diff
@@ -58,6 +58,10 @@ using ElectronPhonon: holstein_model, Structure
         end
     end
 
+    # check_group throws unless the operations form a group (identity, inverses, two-sided
+    # closure) and returns them otherwise; it checks full SymOps, so τ and is_tr count too.
+    is_group(sym) = (ElectronPhonon.check_group(collect(sym)); true)
```
jaemolihm: can't we do `@test ElectronPhonon.check_group(model.symmetry)` ?

src/common/symmetry.jl
```diff
@@ -13,6 +13,7 @@ using LinearAlgebra
 
 export Symmetry
 export symmetry_operations
+export restrict_symmetry_to_dimension
```
jaemolihm: no export

src/models/holstein_model.jl
```diff
@@ -88,9 +66,9 @@ coupling measured against the half-bandwidth. Deriving `g` from `λ` needs `t 
   symmetry group to one that the `dimension`-dimensional Hamiltonian actually obeys. (A cubic
   cell would make spglib report operations mixing `x` and `z`, which is *not* a symmetry of
   e.g. the square-lattice band, and symmetry-reduced calculations would then be wrong.)
-  `model.symmetry` is additionally restricted by `_restrict_symmetry_to_dimension`, so it can
-  never exceed that group regardless of what spglib returns; `model.structure.symmetry` is
-  spglib's own result and may in principle be larger.
+  The symmetry is additionally passed through [`restrict_symmetry_to_dimension`](@ref) (via
```
jaemolihm: no need to explain this detail, just expain the final state

src/models/holstein_model.jl
jaemolihmbot: Done in e1d125f. `julia = "1.11"` on ElectronPhonon, and the gate is gone — `public holstein_model` is now a plain statement.

I also removed the **other** gate, in `src/calculator/AbstractCalculator.jl`. It existed only to keep compat at `"1"` (its comment said so), so leaving it would have left the repo with a dead workaround and a comment asserting something false. That block is now a plain `public ...` list.

MigdalEliashberg.jl and EPSpectral.jl are bumped to `julia = "1.11"` too, but they are **separate git repos**, so those edits are on disk and not part of this PR — they are uncommitted in their own checkouts. MigdalEliashberg.jl also has unrelated uncommitted work in `src/kernel.jl`, `src/observables.jl`, `src/solver.jl` and `test/runtests.jl`, so I did not commit there; say the word and I will commit just the `Project.toml` in each.

test/test_holstein_model.jl
jaemolihmbot: Yes — done in e1d125f. Two things were in the way, both fixed rather than worked around:

- `check_group` returned the operations, so `@test` got a non-Boolean. It now returns `true` and throws otherwise, with a docstring saying so. No caller used the old return value (the four in `test_symmetry.jl` all discard it).
- The `collect` was unnecessary: `check_group` only indexes and iterates, and `test_symmetry.jl` has been passing a `Symmetry` directly since before this branch. My mistake.

So it is now literally `@test ElectronPhonon.check_group(model.symmetry)`, and the local helper is gone.

src/common/symmetry.jl
jaemolihmbot: Done — and then done properly in d7377db: it is neither exported nor `public`, just an internal function reached as `ElectronPhonon.restrict_symmetry_to_dimension`. (e1d125f had switched it to `public`; your follow-up was right that it does not need to be part of the supported surface at all.) The test imports it explicitly, which works regardless.

src/models/holstein_model.jl
jaemolihmbot: Done in e1d125f — the bullet now states the resulting arrangement and drops the reasoning:

```
- `dimension = 3` : spatial dimension of the *hopping*, 1, 2 or 3, recorded as
  `model.dimension`. The `Model` always lives on a 3d lattice: for `dimension < 3` the
  directions without hopping get a lattice constant of `100 * alat`, and `model.symmetry`
  holds only the operations that do not mix them with the hopping directions. `model.volume`
  is therefore `100^(3-dimension) * alat^3`, so per-volume quantities
  (`ElectronOccupationParams`, transport `volume`) must be normalized by the user to a
  length/area if a `dimension < 3` density is wanted.
```

12 lines down to 7, and the "a cubic cell would make spglib report x-z mixing" explanation is gone.

jaemolihm (approved):
Will merge once MigdalEliashberg.jl MCI stuff are finalized

PR #35, opened 2026-09-18
Add alloc_zeros and to_device_copy, the writable-buffer allocators, and collapse the two open-codings

jaemolihmbot:
## Motivation

`alloc(backend, T, dims...)` returns `undef` memory, and `to_device(backend, x)` is deliberately
the identity on a `CPUBackend`. Neither can supply a buffer that will be **written to** — a
reduction target must be zeroed first, and a host "copy" that is really the argument gets mutated
under the caller. Both workarounds were already open-coded in this package, and
MigdalEliashberg.jl carries its own local pair for the same reason.

## Change

`src/common/gpu_utils.jl`, beside `alloc`, neither exported:

```julia
alloc_zeros(backend, ::Type{T}, dims...) where {T} = fill!(alloc(backend, T, dims...), zero(T))
to_device_copy(backend, A::AbstractArray)          = copyto!(alloc(backend, eltype(A), size(A)), A)
```

Each is named after the sibling it pairs with — allocation after `alloc`, the move after
`to_device`.

The two in-tree open-codings now call `alloc_zeros`:

| site | was |
|---|---|
| `src/calculator/tiled_device_output.jl:117` (full-resident branch) | `Any[fill!(alloc(backend, FT, t.dims...), zero(FT)) for _ in 1:t.narr]` |
| `src/boltzmann/boltzmann_calculator.jl:174` (`Sₒ`) | `fill!(alloc(backend, FT, n_i, nT), zero(FT))` |

The `undef` sibling 3 lines above the first one (`tiled_device_output.jl:114`, the tiled branch,
which overwrites the whole tile) stays on `alloc` — the contrast in that one `if/else` is what the
pair is for.

**Swept for the rest and found none:** no `copyto!(alloc(` anywhere, no
`to_device(backend, zeros(...))`, and no split `alloc` + `fill!`. Every other `alloc` site (13 in
`run_eph_over_k_and_kq.jl`, 8 in `run_eph_over_q_and_k.jl`) is a scratch buffer that is fully
overwritten, which is what `alloc` is for. One near-match left alone:
`src/common/occupation.jl:143`'s local `_to_device` takes a `gpu_array` **prototype**, not a
backend, so folding it in would change `bte_compute_μ!`'s signature — a separate change.

## Test

`test/test_backend_alloc.jl` (new, wired into `runtests.jl` after `test_gpu.jl`), both backends,
the GPU arm under `CUDA.functional()`: **44 pass**.

* `alloc_zeros` — shape, element type, actually zeroed, and the same array type `alloc` gives.
  Checked for `Float64`, `ComplexF64` and `Int`, all different from `gpu_backend()`'s
  `CuArray{ComplexF64}` prototype, since element-type-from-the-caller is what `alloc` provides.
* `to_device_copy` — `!== A` **on `CPUBackend` too** (the reason it exists; asserted next to
  `to_device(CPUBackend(), A) === A`), writing the copy leaves `A` alone, and the element type is
  the argument's exactly: `ComplexF64`, `Int` and `Complex{Int}` all survive unpromoted.

The two collapsed sites have no direct unit test; they are covered through the calculators that
own them. Those testsets are unchanged by this PR, run against `main` and against this branch:

| testset | `main` | this branch |
|---|---|---|
| `test/boltzmann/test_gpu_boltzmann_calculator.jl` | 63 | **63** |
| `test/boltzmann/test_multigrid_transport.jl` | 97 | **97** |
| `test/test_calculator_contract.jl` | 20 | **20** |

## Unblocks

MigdalEliashberg.jl#8, which has local copies of both helpers (11 `alloc_zeros` call sites and 4
`to_device_copy` ones) and drops them once this merges.


src/common/gpu_utils.jl
```diff
@@ -66,6 +66,26 @@ end
 alloc(::CPUBackend, ::Type{T}, dims...) where {T} = Array{T}(undef, dims...)
 alloc(b::GPUBackend, ::Type{T}, dims...) where {T} = similar(b.proto, T, dims...)
 
+"""
+    alloc_zeros(backend, ::Type{T}, dims...) -> AbstractArray{T}
+
+Zero-filled array of `T` on `backend`. `alloc` hands back `undef` memory, so a buffer that is
+accumulated INTO rather than fully overwritten — a reduction target such as the Boltzmann
```
jaemolihm: don't mention Boltzmann

jaemolihmbot: `1a110b0`. 다운스트림 모듈 이름을 빼고 "a reduction target" 까지만 말하도록 고쳤습니다 — backend 유틸 docstring 이 Boltzmann 을 알 이유가 없습니다.

jaemolihm (approved):

PR #38, opened 2026-09-21
Add to_device for a sparse matrix and a run-to-run reproducible spmm!

jaemolihmbot:
## Motivation

MigdalEliashberg.jl's compressed kernel is being reframed so that the inner→outer star sum is a
sparse product `V_R · S`, `S[f,i] = weight_f[f]·δ(f_to_i[f], i)` (review comment on
jaemolihm/MigdalEliashberg.jl#13). Two generic GPU pieces are needed for that, and by the rule set
on #11 they belong here rather than in the package using EPjl.

## Change

`ext/ElectronPhononCUDAExt.jl`
- `to_device(::GPUBackend, ::SparseMatrixCSC) = CuSparseMatrixCSR(A)`. Without it a sparse matrix
  matches the generic `to_device(::GPUBackend, ::AbstractArray)` and is **silently densified** —
  894 MB at nk=96, 9.8 TB at nk=144 for the star matrix. This method is not an optimization; it is
  what prevents an OOM far from its cause.

`src/common/gpu_utils.jl` + ext
- `spmm!(C, A, B)`: `mul!` on the host, `CUSPARSE_SPMM_CSR_ALG3` on the device. cuSPARSE's default
  algorithm splits a row across thread blocks and combines the partial sums with atomics, so the
  summation order varies run to run. Measured on an A100 at `(5182 × 237136) · (237136 × 30)`:

| algorithm | bitwise stable run-to-run | time/call |
|---|---|---|
| `DEFAULT` / `CSR_ALG1` | no — 6.1% of entries differ, 1.4e-16 rel | 0.159 ms |
| `CSR_ALG2` | no (despite being documented deterministic) | 0.43 ms |
| **`CSR_ALG3`** | **yes** | 0.275 ms |

Anything whose result feeds an iteration count or a run-to-run comparison must use `spmm!` rather
than `mul!`. The two backends are not bitwise equal to each other (~1 ulp, reassociation within a
row); each is reproducible against itself.

## Test

`test/test_backend_alloc.jl`, new `sparse to_device / spmm!` testset: the device matrix is not a
dense `CuArray` and keeps its `nnz`; `spmm!` is bitwise identical over 10 repeats including
freshly uploaded operands; host and device agree to `8·eps`. 67/67 pass, A100, CUDA.jl 6.2.1.

Stacked on #37 (`feat/reclaim-device-memory`); merge that first.


ext/ElectronPhononCUDAExt.jl
```diff
@@ -41,6 +43,18 @@ ElectronPhonon.synchronize(::ElectronPhonon.GPUBackend) = CUDA.synchronize()
 # `[compat] CUDA = "6"` is what pins; 5.x's `reclaim` is a different, GC-less routine.
 ElectronPhonon.reclaim_device_memory(::ElectronPhonon.GPUBackend) = (CUDA.reclaim(); nothing)
 
+# A sparse matrix matches the generic `to_device(::GPUBackend, ::AbstractArray)` below, which would
+# DENSIFY it — 9.8 TB for the inner→outer star matrix at nk=144. CSR is the layout cuSPARSE's SpMM
```
jaemolihm: 1. you should not mention nk=144, it makes no sense in this low level routine. Just remove 9.8TB part.
2. I think this comment should say
`to_device(::GPUBackend, ::AbstractArray)` always densify the given array. Add a specialization for sparse matrices that preserves sparsity on the device.

jaemolihmbot: Both done in `3376838` — your wording, and the grid size is gone:

```julia
# `to_device(::GPUBackend, ::AbstractArray)` always densifies the given array. Add a specialization
# for sparse matrices that preserves sparsity on the device; CSR is what cuSPARSE's SpMM takes.
```

Agreed on the principle: a caller's `nk` has no business in a backend primitive.

jaemolihm (approved):

Issue #34, opened 2026-09-18
symmetry === nothing: dispatch survey, plus three defects it turned up


jaemolihmbot:
Survey of how the symmetry-present / symmetry-absent distinction is handled across the package,
prompted by one instance we just fixed. Read-only; nothing here has been changed.

### The pattern

In PR #32, `find_unfolding_indices(el_i, el_f, symmetry)` had grown a second, `::Nothing`-specific
method that duplicated the whole per-inner-state loop only to swap one lookup. The fix was to push
the `nothing` decision one layer *down*, onto the lookup itself:

```julia
state_index_in_star(s::AbstractBandStates, xk, iband::Integer, ::Nothing) = state_index(s, xk, iband)
```

after which the caller collapses back to a single method. The general shape to look for: **the
distinction resolved at too high a layer** — a duplicated `::Nothing` method, or an
`if symmetry === nothing` branch inside a larger routine — where a one-line dispatch at the bottom
of the call chain removes the duplication instead.

Coverage: all 34 `::Nothing` dispatch slots in `src/`. `ext/` contains no `symmetry` at all.
MigdalEliashberg.jl has **zero** `symmetry === nothing` / `isnothing(symmetry)` branches and zero
`::Nothing` symmetry slots — its null handling is already delegated upstream by dispatch, so there
is nothing to do on the consumer side.

### 1. `unfold_data_map(state, ::Nothing)` — `src/boltzmann/electron_lbte.jl:229` (14 lines)

Duplicates the `indmap` build of the general method at `:192`. Only two things differ: the k fed to
the key (`xk` vs `Sk = is_tr ? -S*xk : S*xk`) and the return (`I(state.n)` vs the sparse
accumulation). The push-down object **already exists** — `identity_symmetry(::Type{FT}=Float64)`
at `src/common/symmetry.jl:320`, currently with a single caller (`src/common/Structure.jl:35`) — so
this is not a matter of inventing a null singleton:

```julia
unfold_data_map(state::BTorBandStates{FT}, ::Nothing) where {FT} =
    unfold_data_map(state, identity_symmetry(FT))
```

**But it is not free, and probably not worth it as stated.** The no-symmetry map's eltype goes
`Bool` → `Mat3{FT}`, which propagates through `:313` into `_solve_bte`'s iteration matrix at
`:127`, i.e. roughly 9× the sparse-map bytes on the *common* full-BTE path (`solve_electron_bte`,
[...]

PR #41, opened 2026-09-21
Fix the three symmetry defects in issue #34: states_index_map star keys, filter_electron_states silently dropping symmetry, unfold_ElectronStates aliasing contract

jaemolihmbot:
Addresses only the **"Defects found on the way"** section (A, B, C) of #34. The dispatch/refactor
survey is out of scope; in particular the shared `_grid_key` helper was **declined by the user**
("this duplication is not worth removing, the code is simple enough to understand"), so all five
open-coded grid-key sites are left as they are.

## A. `states_index_map` star keys were wrong (`src/boltzmann/BTStates.jl`)

Problem: the symmetry block built star keys as `is_tr * S * xk` with `is_tr::Bool`.

- before: `Sk = is_tr * S * xk` → the **zero vector** when `is_tr == false`, `+S*xk` (no minus) when true
- after: `Sk = apply_symop(symop, xk, :momentum)`, iterating the `Symmetry` as `SymOp`s

Second problem, same routine: the `xk_shift` keyword. It had **no call site anywhere** in
ElectronPhonon.jl, MigdalEliashberg.jl, EPSpectral.jl or the `broadening/` and `superconductivity/`
driver dirs, and it is ill-defined together with `symmetry` in the first place (a shifted grid
`{n/N + s}` is invariant only under the subgroup with `S*s == s (mod 1/N)`, so a star key for a
shifted state is either a silent `round` snap to a neighbouring grid point or the key of a point
outside the star). Guarding a combination of two unused features is not worth the lines, so the
keyword is **removed** (`a27509a`, reverting the guard that `6c2ac73` added). The key build goes
back to the plain grid-key recipe, `mod.(round.(Int, states.xks[i] .* states.ngrid), states.ngrid)`.

The three callers pass a single positional argument and are untouched.

**No existing numerical result changes.** Re-verified at `fe12845`: all three in-package call sites
pass a single argument (`screening_rpa.jl:99`, `boltzmann/electron_master_equation.jl:26`,
`boltzmann/electron_serta.jl:302`), no test passes a symmetry, and `xk_shift` has zero callers
anywhere in the EPjl stack. Per the issue's suggestion to delete the parameter: the user chose to
**keep** `symmetry` and fix the formula; `xk_shift` is gone, so the whole diff against `main` in
this file is the `apply_symop` star key plus the docstring that documents it.

## B. `_filter_with_band_ranges` dropping `symmetry` is a contract, not a bug (`src/filter.jl`)

`symmetry` is used only on the `kpts_input isa NTuple{3,Integer}` branch; the else branch takes the
prebuilt k-set as given. #34 read that as a silent drop. It is not, and the first revision of this
PR got it wrong by throwing an `ArgumentError` there. Two reasons the throw was wrong:

- `run_eph_over_k_and_kq` declares both positionals as
  `Union{NTuple{3,Int}, Kpoints, GridKpoints, FilteredBandStates}`
  (`src/calculator/run_eph_over_k_and_kq.jl:36-37`) and `_setup_electron_k` / `_setup_electron_kq`
  forward `symmetry` unconditionally, so a prebuilt k-set plus a symmetry is admitted by the
  documented API regardless of what today's in-repo call sites happen to pass.
- The combination is *meaningful*. `_setup_electron_kq` branch (2) passes the filter result
  straight to `unfold_kpoints(sel_irr.kpts, symmetry)` (`src/calculator/eph_setup_common.jl:106`):
  a caller that has already reduced to the IBZ hands in that set, needs it back untouched, and
  still wants `symmetry` live for the unfold. The throw broke exactly that workflow.

So `src/filter.jl` is **documentation-only relative to `main`**. The contract is now explicit on
the public `filter_electron_states` docstring — `symmetry` reduces only on the `NTuple` grid path,
where `kpoints_grid` builds the irreducible set; a prebuilt `Kpoints`/`GridKpoints` is taken as
given and not reduced, deliberately — with a short version kept at the else branch so the next
reader does not re-file it as defect B again.

The `shift` guard immediately above (`error("nonzero shift and symmetry incompatible (not
implemented)")`, `filter.jl:173-174`) is untouched: it guards the grid path, where the reduction
really does happen.

## C. `unfold_ElectronStates` aliasing contract (`src/boltzmann/unfold.jl`)

Docstring only, no behavior change: document that a trivial symmetry (`nothing` or `nsym == 1`)
returns `states_irr` itself rather than a copy (a `deepcopy` would double a large per-k state
array), and cite the caller that already guards on identity,
`el_kq_save_irr !== el_kq_save && empty!(el_kq_save_irr)` in `eph_setup_common.jl`.
The `# FIXME: Add test` stays, since no test is added.

## Test

- **New** `test/boltzmann/test_BTStates.jl` (+8 assertions, wired into `runtests.jl`) — the symmetry
  path of `states_index_map`, which had no coverage at all. Model-free, an IBZ-reduced 4³ grid,
  two point groups because they constrain different halves of the transform:
  * fcc, `Oh` + TR (96 ops) — pins the rotation part
  * zincblende, `Td` + TR (48 ops) — pins the time-reversal sign, whose star points no rotation reaches

  Teeth, measured by re-running the old `is_tr * S * xk` formula in the same session:

  | group | wrong star lookups (old) | keys absent (old) | grid points covered (old / want) |
  |---|---|---|---|
  | fcc `Oh` | 192 / 1536 | 0 | 64 / 64 |
  | zincblende `Td` | 192 / 768 | 96 | **48 / 64** |

- `test/test_filter_electron_states.jl` +9 (15 → 24): the contract pinned both ways. A prebuilt
  `Kpoints` and a prebuilt `GridKpoints`, each with and without `symmetry`, give the same
  selection (`kpts.n`, the `(xk, iband)` state set, `nstates_base`, `band_range`); and the same
  symmetry on the `NTuple` grid path *does* reduce `kpts.n`, so the equality is not vacuous.
- Full suite `Pkg.test("ElectronPhonon")` on an RTX A6000 (GPU testsets included):
  **23051 pass, 0 fail**, 12m46s at `6f955df`, against a base of 23039. The three follow-up
  commits were checked with targeted runs rather than another full suite:
  `75397bf` (the defect-B contract; comments, a docstring and its test only) with
  `test/test_filter_electron_states.jl` 24/24; `a27509a` (the `xk_shift` removal) with
  `test/boltzmann/test_BTStates.jl` 8/8 and `test/test_screening.jl` 10/10 (the one
  `states_index_map` consumer with live coverage), in a **restarted** session so no stale
  Revise method could hide the removed keyword — `states_index_map(st; xk_shift = ...)` is now a
  `MethodError`. Net added assertions: 8 + 9 = 17.


src/filter.jl
```diff
@@ -198,10 +202,12 @@ range, `sel.nstates_base` the below-window carrier count, `state_weights(sel)` t
 per-k) BZ weights. Supersedes the legacy tuple-returning `filter_kpoints`.
 
 `shift` applies only to the `NTuple` grid path (ignored for a prebuilt `Kpoints`/`GridKpoints`) and
-is mutually exclusive with `symmetry`. Under `mpi_comm` the k-points are split across ranks: the grid
-is built distributed, each rank filters its slice once (single eigensolve pass), then the kept
-k-points and per-k band ranges are redistributed together through the same gather/scatter so they
-stay aligned, and the local below-window counts are summed.
+is mutually exclusive with `symmetry`. `symmetry` likewise requires the `NTuple` path, since the
+irreducible k-set is built by `kpoints_grid`: a prebuilt `Kpoints`/`GridKpoints` passed together
+with a `symmetry` is an `ArgumentError`, not a silent full-BZ selection. Under `mpi_comm` the
```
jaemolihm: in run_eph, don't we pass k_input to filter, and this will block that? Let's make symmetry + Kpoints -> no reduction a contract.

src/boltzmann/BTStates.jl
```diff
@@ -125,8 +125,13 @@ end
 # TODO: Make ElectronState and PhononState similar so only one function is needed.
 
 """
-    states_index_map(states, symmetry=nothing)
+    states_index_map(states, symmetry=nothing; xk_shift=Vec3(0, 0, 0))
```
jaemolihm: what is the usecase for xk_shift ?

jaemolihmbot: There is none. `xk_shift` has exactly two occurrences in the whole tree, and both are the definition itself — the signature and the one use inside the body:

```
src/boltzmann/BTStates.jl:131:function states_index_map(states, symmetry=nothing; xk_shift=Vec3(0, 0, 0))
src/boltzmann/BTStates.jl:136:        xk_int = mod.(round.(Int, (states.xks[i] - xk_shift) .* states.ngrid), states.ngrid)
```

Zero call sites across ElectronPhonon.jl, MigdalEliashberg.jl, EPSpectral.jl and the `broadening/` and `superconductivity/` driver dirs. The three in-package callers of `states_index_map` (`screening_rpa.jl:99`, `boltzmann/electron_master_equation.jl:26`, `boltzmann/electron_serta.jl:302`) all pass a single positional argument, so neither the keyword nor the symmetry path is exercised by anything.

So the guard this PR adds is a guard on a combination of two unused features. If you would rather drop the keyword than document it, that is a strictly smaller diff — it deletes the keyword, the guard, and the guard test, and leaves `states_index_map(states, symmetry)` doing the one thing it is actually used for. The only cost is that it is a public keyword, so an off-tree script could in principle be passing it. Your call; I left it in because you chose the guard over removal earlier, before this had come up.

src/filter.jl
jaemolihmbot: Agreed, and done in 75397bf. The `ArgumentError` is gone; `src/filter.jl` is now documentation-only relative to `main`.

You are right about `run_eph`: `run_eph_over_k_and_kq` declares both positionals as `Union{NTuple{3,Int}, Kpoints, GridKpoints, FilteredBandStates}` and `_setup_electron_k` / `_setup_electron_kq` forward `symmetry` unconditionally, so the combination is admitted by the documented API. My audit only established that no *current in-repo* call site reaches it, which was the wrong test. It is also a meaningful combination rather than a user error: `_setup_electron_kq` branch (2) passes the filter result straight to `unfold_kpoints(sel_irr.kpts, symmetry)`, so a caller that has already reduced to the IBZ needs its set back untouched with the symmetry still live for the unfold.

The contract is now stated on the public `filter_electron_states` docstring, with a short version at the else branch so the next reader does not re-file it as defect B. The test pins it both ways: a prebuilt `Kpoints` and a prebuilt `GridKpoints`, each with and without `symmetry`, give the same selection (`kpts.n`, the `(xk, iband)` state set, `nstates_base`, `band_range`); and the same symmetry on the `NTuple` grid path does reduce `kpts.n`, so the equality is not vacuous. `test/test_filter_electron_states.jl` 24/24.

The `shift` guard just above is untouched: it guards the grid path, where the reduction really does happen.

src/boltzmann/BTStates.jl
jaemolihm: delete xk_shift argument

jaemolihmbot: Done in a27509a.

- `xk_shift` deleted from `states_index_map`; the key build is back to the plain grid-key recipe, `mod.(round.(Int, states.xks[i] .* states.ngrid), states.ngrid)`.
- The guard from 6c2ac73 and its two tests go with it: it only existed to reject `xk_shift` together with `symmetry`, and a guard on a parameter that no longer exists would be worse than never having added it.
- Docstring now describes just the map and the symmetry-star keys.

Verified: no occurrence of `xk_shift` remains in ElectronPhonon.jl, MigdalEliashberg.jl, EPSpectral.jl or the `broadening/` and `superconductivity/` driver dirs. The three callers (`screening_rpa.jl:99`, `boltzmann/electron_master_equation.jl:26`, `boltzmann/electron_serta.jl:302`) pass a single positional argument and are untouched.

Ran in a **restarted** session, so no stale Revise method could hide the removal: `test/boltzmann/test_BTStates.jl` 8/8, `test/test_screening.jl` 10/10 (the one consumer with live coverage, via `screening_rpa.jl:99`), and `states_index_map(st; xk_shift = ...)` is now a `MethodError`.

The remaining diff against `main` in this file is defect A alone: `apply_symop(symop, xk, :momentum)` in place of `is_tr * S * xk`, plus the docstring sentence that documents the star keys. The key expression is now identical to the four other open-coded grid-key sites, and per your earlier call it is left as the literal expression rather than factored into a shared helper.

jaemolihm (approved):

PR #32, opened 2026-09-18
Complete the serial (comm === nothing) path, and unfold without symmetry

jaemolihmbot:
## Motivation

MigdalEliashberg.jl wants to write one code path whether or not MPI and symmetry are in
play, and two things here stopped it: `find_unfolding_indices` had no method for "no
symmetry group", and roughly half the `mpi_*` helpers had no `comm === nothing` method,
so `nothing` worked for some of them and `MethodError`d for others.

## Change

**`find_unfolding_indices` is one method, on the grid hash** (`src/band_states.jl`)

It looks the inner state's `(k, band)` up in the outer set instead of scanning it inside
the symmetry loop:

```julia
j = state_index_in_star(el_i, xk_f, ib, symmetry)
```

Searching the star of the INNER k finds the same state as applying every operation to
every OUTER k, because the momentum actions form a group: `∃S: S·k_j ≡ k_f` iff
`∃S: k_j ≡ S·k_f`. An inner state therefore costs `nsym` O(1) lookups rather than a scan
of `el_i` — O(nsym·n_f) instead of O(nsym·n_f·n_i). An exact `k_i ≡ k_f` hit wins over a
star image, so the answer is unique exactly when `el_i` holds one member of each star,
which is what an IBZ reduction gives it. `el_i` must carry `GridKpoints` (the lookup is
its integer-grid hash); that shows up as `state_index`'s `MethodError`.

`symmetry === nothing` gets **no method of its own** — the star of `k_f` is `{k_f}`, so it
belongs one layer down:

```julia
state_index_in_star(s::AbstractBandStates, xk, iband::Integer, ::Nothing) =
    state_index(s, xk, Int(iband))
```

`state_indices_full_star` had the same gap and gets the same member (that one index, or
empty). The return of `find_unfolding_indices` is now **`Vector{Int}`**, because no caller
of the `BandStates` method reads the symmetry-operation index (the `BTStates` method keeps
returning it — there it unfolds a velocity).

**`::Nothing` methods for every remaining `mpi_*` helper** (`src/common/mpi.jl`)

| had one already | added |
|---|---|
[...]

PR #42, opened 2026-09-21
feat: gather_global_states, the rank-ordered global BandStates gather

### LINKED ISSUE #36: Move MigdalEliashberg's _gather_global_outer (rank-ordered global BandStates gather) into EP
MigdalEliashberg.jl#5 리뷰에서 나왔습니다 (`src/kernel.jl:317`, *"seems this belong to EPjl. Keep for now but open issue in EPjl"*). 그 PR 에서는 ME 에 그대로 두고, 이 이슈로 남깁니다.

ME 가 들고 있는 함수:

```julia
# Gather the GLOBAL outer states across `mpi_comm`, in rank order (rank 0's states first, then
# rank 1, …) — the same order the eph k-split produces, so this rank's `states_i.n` local states
# occupy the contiguous global block `i_offset+1 : i_offset+states_i.n`. Returns
# `(gstates, i_offset, i_counts)`. Serial (`comm === nothing`): `states_i` itself, `i_offset = 0`.
function _gather_global_outer(states_i::BandStates, mpi_comm)
    mpi_comm === nothing && return states_i, 0, [states_i.n]
    counts = ElectronPhonon.mpi_allgather([states_i.n], mpi_comm)
    i_offset = sum(@view counts[1:ElectronPhonon.mpi_myrank(mpi_comm)])
    ElectronPhonon.mpi_allgather(states_i, mpi_comm), i_offset, Vector{Int}(counts)
end
```

EP 로 올라가야 할 이유:

- **ME 고유한 것이 하나도 없습니다.** `BandStates` 와 communicator 만 알면 되고, 반환하는 세 값은 전부 `BandStates` 의 랭크 분할에 대한 사실입니다. 전자-포논이나 Eliashberg 와 무관합니다.
- **조각들이 이미 EP 에 있습니다** — `mpi_allgather(::BandStates, ::MPI.Comm)`(`test/test_band_states.jl:84` 에서 `COMM_SELF` 로 테스트됨), `mpi_allgather`, `mpi_myrank`, 그리고 `::Nothing` 직렬 경로(#32). 이 함수는 그 넷의 얇은 조합입니다.
- **오프셋 규약이 EP 의 것입니다.** "랭크 순서 연접, 로컬 블록이 `i_offset+1 : i_offset+n_local`" 은 EP 의 eph k-split 이 만드는 순서이고, ME 는 그걸 **가정**하고 있을 뿐입니다. 규약을 만드는 쪽에 규약을 읽는 코드가 있는 편이 맞습니다. 이 가정이 깨지면 조용히 틀린 `f_to_i` 가 나옵니다 — ME 쪽 주석이 그래서 그 순서를 명시하고 있습니다.

다른 소비자가 있을 법한 곳: 랭크 분할된 `BandStates` 를 들고 글로벌 인덱스로 무언가를 조회해야 하는 코드라면 같은 셋이 필요합니다.

제안: `src/band_states.jl` 의 MPI 헬퍼 옆에 `gather_global_states(states, comm) -> (gstates, offset, counts)` 같은 이름으로. export 는 그 파일의 관례대로 하지 않고, 직렬 경로(`comm === nothing`)는 지금처럼 입력을 그대로 돌려주면 됩니다 — 다만 그 경우 **반환이 입력과 같은 객체**라는 점은 docstring 에 명시되어야 합니다 (#33 1번에서 `to_device` 에 대해 지적된 것과 같은 aliasing 계약).

jaemolihmbot:
Moves MigdalEliashberg's `_gather_global_outer` into `src/band_states.jl` as
`gather_band_states(states, comm) -> (gstates, offset, counts)` (not exported).

Nothing in it was ME-specific: it composes `mpi_allgather(::BandStates, ::MPI.Comm)`,
`mpi_allgather`, `mpi_myrank` and their `::Nothing` serial methods, and the three values it
returns are all facts about EP's own rank split. The offset convention - rank-ordered
concatenation, so the local states are the global block `offset+1 : offset+n` - is made by
EP's k-splitters; ME was only assuming it, and a broken assumption gives a silently wrong
index map. The docstring now states both that convention and the aliasing contract
(serially, `gstates` IS the input object, not a copy).

The serial path needs no branch: `mpi_allgather(x, ::Nothing) = x` and
`mpi_myrank(::Nothing) = 0` (#32) already give exactly the behaviour ME hand-coded.

Test in `test/test_band_states.jl` next to the existing `mpi_allgather` one: `COMM_SELF`
(offset/counts bookkeeping and the block convention) plus the serial path, including
`gstates === states` and `offset == 0`. 93 -> 99 passes in that file. The branch name predates the rename.

`Pkg.test("ElectronPhonon")` 23045/23045. The ME side is jaemolihm/MigdalEliashberg.jl#14,
which must land after this.

Closes #36.


src/band_states.jl
```diff
@@ -329,6 +329,26 @@ function mpi_allgather(s::BandStates{FT}, comm::MPI.Comm) where {FT}
...
+rank 0's states first, then rank 1's, and so on, which is the order the e-ph k-split produces.
+This rank's `states.n` local states are therefore the contiguous global block
+`offset+1 : offset+states.n`. Anything that indexes a gathered set by a global index depends on
+that ordering; it is EP's convention, made by the k-splitters, so it is read here rather than
+assumed by the caller.
+
+Serially (`comm === nothing`) the returned `gstates` IS `states`, the same object and not a copy,
+so a caller that mutates it mutates its input.
+"""
+function gather_global_states(states::BandStates, comm)
```
jaemolihm: why not mpi_allgather(::BandStates, comm) ?

jaemolihmbot: Because `mpi_allgather(::BandStates, comm)` returns only the gathered `BandStates` — it does not give you the offset or the per-rank counts, and those are the parts issue #36 said belong in EP. The gather itself was already here; what ME was carrying was the bookkeeping around it.

Folding them into `mpi_allgather` would break two things:

1. **The family contract.** Every `mpi_allgather` method returns the gathered object and nothing else (`src/common/mpi.jl:118`, `src/band_states.jl:318`). Returning a 3-tuple from the `BandStates` method alone would make it the odd one out, and would break its existing single-return caller and test.
2. **The serial path, specifically.** `gather_global_states` has no `comm === nothing` branch — it works serially only because the generic `mpi_allgather(x, ::Nothing) = x` (`src/common/mpi.jl:128`) and `mpi_myrank(::Nothing) = 0` already produce the right triple. If the `BandStates` method returned a tuple, that generic `::Nothing` fallback would no longer have the matching shape and the serial path would need a hand-written branch back.

Worth noting the two are closer than they look: `mpi_allgather(::BandStates, …)` already does this exact rank-prefix-sum internally at `:320-322`, but over `kpts.n` to renumber `iks`, whereas `gather_global_states` needs it over `states.n` for the state index space `f_to_i` lives in. Same shape, different quantity — which is an argument for them being two functions rather than one.

If your point is that the 3-line wrapper does not earn a name, the alternative is putting those three lines back at the ME call site, which is what #36 asked to remove. Happy to go either way — say which and I will change it.

jaemolihm: what about gather_band_states ?

jaemolihmbot: Renamed to `gather_band_states` in 7162d22, and the ME side follows in jaemolihm/MigdalEliashberg.jl@42ecda0.

The docstring prose keeps "global" where it describes what `gstates` and `offset` mean; only the name changed. Checked: no `gather_global_states` left in either repo.

Re-ran `test/test_band_states.jl` (99/99) and, on the ME branch, the "GPU EliashbergCalculator (full-band + windowed)" testset (9/9), which is the one that builds an `EliashbergKernel` from a calculator and so goes through the renamed call site.

jaemolihm (approved):

src/band_states.jl
jaemolihmbot: Follow-up: the ME side was rebased onto ME `main` (7d2b98e), so that hash is now jaemolihm/MigdalEliashberg.jl@b557728. The branch's contribution is byte-identical to the pre-rebase one; only the base moved.

`Pkg.test("MigdalEliashberg")` on the rebased tree: 283 pass / 0 fail / 0 error, GPU testsets included. That run used EP `main` (16a3a2d) plus this PR's two commits, because ME `main` now imports `reclaim_device_memory`, which landed in #37 - after this branch's base `fe12845`. This PR is unaffected and still merges cleanly.

